PR #91, opened 2024-06-02
feat: show progress while translating

vrcprl:
+ added more data to .json file to see the progress

### DIFF
--- a/packages/cli/src/services/bucket/base.ts
+++ b/packages/cli/src/services/bucket/base.ts
@@ -2,6 +2,8 @@ import _ from "lodash";
 import fs from 'fs';
 import { BucketPayload, BucketTranslatorFn, IBucketProcessor } from "./core.js";
 
+const CHUNK_SIZE = 25;
+
 export abstract class BaseBucketProcessor implements IBucketProcessor {
   constructor(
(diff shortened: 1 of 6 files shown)


packages/cli/src/services/bucket/base.ts
```diff
@@ -24,20 +26,28 @@ export abstract class BaseBucketProcessor implements IBucketProcessor {
...
       const partialData = _.pick(payload.data, batch);
       const partialPayload = { data: partialData, meta: payload.meta };
       const partialResult = await this.translator(sourceLocale, targetLocale, partialPayload);
       _.merge(resultData, partialResult);
+      completedBatches++;
+
+      if (onProgress) {
+        const progress = (completedBatches / totalBatches) * 100;
+        onProgress(progress);
+      }
```
maxprilutskiy: nice 👍

maxprilutskiy: Actually you can even change it to

```suggestion
      const progress = (completedBatches / totalBatches) * 100;
      onProgress?.(progress);
```

maxprilutskiy (approved):

PR #158, opened 2024-08-19
feat(spec): exclusion patterns

### LINKED ISSUE #151: Exclude files from a bucket
### What

This is a feature request.
We would like to be able to exclude some paths / documents from a bucket.

### Why

We will use replexica to translate blog posts from one language to another. We want 95% of the blog-posts to be available in both languages. But for 5%, this makes no sense. One blog post can be very specific for one country for example.

### How

Here is how graphql-codgen which also has a bucket concept handle it :
image

Have you noticed the "!" in front of the paths to exclude ?

maxprilutskiy:
Major upgrade:

`i18n.json` now supports file exclusions. Here's how:

```json
"buckets": {
  "markdown": {
    "include": [
      "demo/markdown-multi/[locale]/*.md",
      "demo/markdown/[locale].md"
    ],
    "exclude": [
      "demo/markdown-multi/[locale]/ignored-blog-post.md"
    ]
  }
},
```

Key changes:

1. Bucket type (i.e. `markdown`, `json`, etc) is now the *key* in the `buckets` object map.
2. Each bucket configuration node has two fields: `include` (required) and `exclude` (optional).

Example: Despite `ignored-blog-post.md` matching the `demo/markdown-multi/[locale]/*.md` pattern, `replexica i18n` ignores it due to the `exclude` field entry.

--

Resolves #151

### DIFF
--- a/.changeset/honest-keys-deliver.md
+++ b/.changeset/honest-keys-deliver.md
@@ -1,0 +1,9 @@
+---
+"@replexica/compiler": minor
+"@replexica/react": minor
+"@replexica/spec": minor
+"@replexica/cli": minor
+"replexica": minor
+---
+
+Configuration spec v1.1: Improved bucket config structure, to support exclusion patterns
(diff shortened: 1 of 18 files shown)


quentin-decre:
You are great ! So fast !

Issue #198, opened 2024-09-08
Fix CLI overwriting i18n.json when unsupported locale is added


maxprilutskiy:
### What
The CLI overwrites i18n.json with a default template when an unsupported locale is added. It should throw an error instead.

### Why
Overwriting the file can lead to data loss and confusion. Users should be alerted to their mistake immediately.

### How
1. Modify the CLI to check for unsupported locales before processing i18n.json.
2. If an unsupported locale is found, throw a clear error message.
3. Ensure the original i18n.json remains untouched.
4. Update documentation to clarify supported locales and error handling.

dankgarlic1:
/attempt
I am unable to authenticate via cli

PR #245, opened 2024-10-16
fix: prevent overwriting i18n.json for unsupported locales

Nithishvb:
Fixes: #198 

This PR modifies the CLI to prevent overwriting i18n.json with a default template when an unsupported locale is provided. Instead of allowing the file to be overwritten, the CLI now checks for unsupported locales and throws a clear error message if one is detected.

### DIFF
--- a/packages/spec/src/config.spec.ts
+++ b/packages/spec/src/config.spec.ts
@@ -48,6 +48,18 @@ const createV1_2Config = () => ({
   version: 1.2,
 });
 
+const createInvalidLocaleConfig = () => ({
+  version: 1,
+  locale: {
+    source: 'bbbb',
+    targets: ['es', 'aaaa'],
+  },
+  buckets: {
+    'src/ui/[locale]/.json': 'json',
+    'src/blog/[locale]/*.md': 'markdown',
+  },
+});
+
 describe('I18n Config Parser', () => {
   it('should upgrade v0 config to v1.2', () => {
     const v0Config = createV0Config();
@@ -106,5 +118,11 @@ describe('I18n Config Parser', () => {
     
     expect(result).not.toHaveProperty('extraField');
     expect(result).toEqual(createV1_1Config());
+  });
+
+  it('should throw an error for unsupported locales', () => {
+    const invalidLocaleConfig = createInvalidLocaleConfig();
+    expect(() => parseI18nConfig(invalidLocaleConfig))
+      .toThrow(`\nUnsupported locale: ${invalidLocaleConfig.locale.source}\nUnsupported locale: ${invalidLocaleConfig.locale.targets[1]}`);
   });
 });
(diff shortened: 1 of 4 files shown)


maxprilutskiy:
Hi @Nithishvb – good attempt!

However, `resave` must remain `true` here: we have a mechanism, that automatically migrates config file to the next version when there's such, whenever `i18n` command is being run.

Here's more info: https://github.com/replexica/replexica/blob/main/packages/spec/src/config.ts

I think the issue might be somewhere in `extendConfigDefinition` function, but that's just a guess.

Also, would be awesome to have a super tiny vitest test, to make sure the fix actually works!

Wanna try again @Nithishvb ?

Nithishvb:
Hi @maxprilutskiy , 

Thanks for the feedback! I'll look into the `extendConfigDefinition` function and ensure resave remains true. I'll also add a vitest test to verify the fix. Let me give it another go!

Nithishvb:
Hi @maxprilutskiy 

I’ve made the requested changes!

1. Resave remains true as per the migration mechanism, which automatically updates the config file to the next version whenever the i18n command runs. 
2. Also, I’ve added a tiny vitest test to ensure the fix works correctly, covering the unsupported locale scenario.

packages/spec/src/config.spec.ts
```diff
@@ -48,6 +48,18 @@ const createV1_2Config = () => ({
...
+  locale: {
+    source: 'bbbb',
+    targets: ['es', 'aaaa'],
+  },
+  buckets: {
+    'src/ui/[locale]/.json': 'json',
+    'src/blog/[locale]/*.md': 'markdown',
+  },
+});
+
```
maxprilutskiy, line 62: nice, thanks for adding the test!

packages/spec/src/config.ts
```diff
@@ -45,6 +45,31 @@ const extendConfigDefinition = <T extends Z.ZodRawShape, P extends Z.ZodRawShape
...
+        }
+    
+        return `Unsupported locale: ${unsupportedLocale}`;
+      });
+    
+
+      if (localeErrors.length > 0) {
+        throw new Error(`\n${localeErrors.join('\n')}`);
+      }
+
```
maxprilutskiy, line 72: 👏

maxprilutskiy (changes requested):
Looks good!

Now, to make the CI/CD checks pass, please:

1. Go to the root of the repo
2. Run `pnpm new`. Select `replexica` AND `replexica/cli`. Then, make sure you trigger the patch version release. (`x.y.Z` <-).
3. Choose a message for your changelog entry.
4. Commit the file that got created, and push it to github.

Looking forward to merging your PR soon!

Nithishvb:
hey @maxprilutskiy 

I've completed the requested changes and pushed them to GitHub. Looking forward to your review and merging the PR soon!

maxprilutskiy (approved):
LGTM!

Issue #192, opened 2024-08-28
Action must do nothing when commit message contains '[skip i18n]'


maxprilutskiy:
### What
Add a feature to our custom GitHub Action that skips i18n processing when the commit message contains '[skip i18n]'.

### Why
This allows developers to bypass i18n checks, if that's needed for debugging GitHub workflows.

### How
1. Parse the current commit message in the action
2. Check for '[skip i18n]' string (case-insensitive)
3. If found, exit the action early without processing, and print a message
4. If not found, continue with normal i18n checks

We could use a regex pattern to match the string, accounting for possible variations in spacing or capitalization.

Dhruvnotfound:
hey would love to work on this!

maxprilutskiy:
reopened, more details here: https://github.com/replexica/replexica/pull/249#issuecomment-2425187681

partik03:
@maxprilutskiy This issue should be closed as already done

PR #249, opened 2024-10-18
feat: add ability to skip i18n processing when commit message contain…

Nithishvb:
Fixes: #192

This PR introduces a new feature in custom GitHub Action that allows developers to bypass i18n processing by including the string `'[skip i18n]'` in their commit message. This is particularly useful for debugging GitHub workflows, enabling teams to quickly skip i18n checks as needed.

### DIFF
--- a/action/entrypoint.sh
+++ b/action/entrypoint.sh
@@ -101,6 +101,15 @@ create_or_update_pr() {
     fi
 }
 
+check_skip_i18n() {
+    local commit_message="$REPLEXICA_COMMIT_MESSAGE"
+
+    if [[ "$commit_message" =~ \[\ *[sS][kK][iI][pP]\ +[iI]18[nN]\ *\] ]]; then
+        echo "::notice::i18n processing has been skipped due to '[skip i18n]' found in the commit message."
+        exit 0
+    fi
+}
+
 # Main execution
 main() {
     # Configure git for committing changes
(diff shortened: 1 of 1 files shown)


Dhruvnotfound:
instead of this implementing the skip directly in the pipeline should be better i think?

maxprilutskiy:
@Nithishvb @Dhruvnotfound both options are fine! The less the user must configure manually, the better.

Thanks everyone! Great job!

maxprilutskiy:
@Nithishvb I'm sorry to say that, but:

https://github.com/replexica/demo/actions/runs/11428979934/job/31794840562

I think it's not working, because we're using REPLEXICA_COMMIT_MESSAGE to look for the skip 18n tag, which is a commit that integration creates. Instead, we should be looking at the message of the commit that triggers the workflow.

Keep in mind, that, theoretically, workflow might be triggered not by push to a branch, but also manually through workflow dispatch, or by new tag event or label, so the commit message env variable that GitHub itself gives access to might simply be empty. If it's empty - we should not skip anything, but just FYI that this use case exists.

Nithishvb:
Hi @maxprilutskiy 

Thanks for your feedback! How about this approach: I’ll modify the function to use `git log -1 --pretty=%B` to check the latest commit message instead of REPLEXICA_COMMIT_MESSAGE. This should ensure we’re looking at the correct commit.
Let me know if you have any other suggestions!

PR #253, opened 2024-10-20
#192 feat:Action skip on commit message contains '[skip i18n]'

Dhruvnotfound:
This PR introduces a new feature in custom GitHub Action that allows developers to bypass Github action workflows by including the string '[skip i18n]' in their commit message, enabling teams to quickly skip i18n checks as needed.

This pr adds the skip to the pr-check and release workflows tell me if u want to do that for the pr-lint workflow as well ;D

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -17,6 +17,12 @@ jobs:
       contents: write
       pull-requests: write
     steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


maxprilutskiy:
hey @Dhruvnotfound thanks for the PR, great job!

We need this in our repo too, so I'm accepting the PR, but ideally this feature should be a part of tha GHA, so that the i18n could be skipped on the user side without requiring the user to configure GH workflows manually.

Issue #197, opened 2024-09-05
Respect existing i18n.json formatting to avoid unnecessary PRs


maxprilutskiy:
### What
Our CLI's JSON serializer, that handles reads/writes of the `i18n.json`, overwrites `i18n.json` with a default format, ignoring existing formatting from tools like Prettier, that user's repo might have.

This triggers useless PRs with only format changes when used with GitHub Actions. Example: https://github.com/calcom/cal.com/pull/16505/files

Originally, we fixed that in #195  by not rewriting `i18n.json` when there's no meaningful changes made in it. However, the update still happens when @lingodotdev team upgrades the `i18n.json` structure design, and that breaks the formatting.

Replexica CLI must respect the style guide of the user's repo.

### Why
- Prevents noise in PRs and commit history
- Respects user's existing code style choices
- Improves compatibility with popular formatting tools
- Reduces confusion and manual work for users

### How
1. Detect existing formatting in i18n.json (indentation, quote style, etc.)
2. Update our JSON serializer to preserve detected formatting
3. If no clear format is detected, fall back to our default
4. Test with popular formatting tools (Prettier, EditorConfig, etc.)
5. Update docs to mention this behavior

Aditya-A-G:
hey, I would like to work on this issue.

partik03:
@maxprilutskiy working on it using prettier to resolve the format but the issue is the function is an async one so should I introduce promises to handle if it's okay with the project?

kapir:
This would be great if it could be fixed, 
We use prettier, and i suspect most in the js ecossystem do.

PR #280, opened 2024-11-01
fix: preserve i18n.json formatting to avoid unnecessary PRs

Aditya-A-G:
**fixes**: #197 

This PR updates the CLI's JSON serializer to detect and respect existing indentation in i18n.json, avoiding unnecessary formatting changes, addressing the core need outlined in #197. Other formatting elements like quote style aren’t needed for this solution.


Aditya-A-G:
Hey @maxprilutskiy, could you review this PR when you have a moment? Thanks!

maxprilutskiy:
hey @Aditya-A-G – thanks for the PR!

Actually, given some recent changes we made, you can now implement #197 in a much more elegant / easier way.

1. We already have `prettier` among dependecies of this project
2. Prettier has a JS API
3. Upon (re-)saving the `i18n.json` configuration, let's use the prettier config the user *might* already have, and if they don't then let's use the default prettier config. (Our only requirement to the the default fallback is 2 spaces indentation).

Once you finish the implementation, don't forget to run `pnpm new` from the root of the repo, and then add a brief summary of the changes you're making. Your message will then trigger a new npm package release, once the PR is merged, and will become a part of the changelog.

Let me know if this makes sense!

Aditya-A-G:
Yes, that makes sense! I will implement the change using Prettier as you suggested.

maxprilutskiy:
@Aditya-A-G let me know how this goes and if you need any help!

cc @partik03 if you have any opinion here, lmk

partik03:
@maxprilutskiy I tried searching for this using prettier but ``prettier.resolveConfig`` function is an asynchronous function which needs to be awaited to resolve the configuration of the i18n file due to which we will have to handle promises at various instances in the code

edit: check the prettier doc [here](https://prettier.io/docs/en/api.html#prettierresolveconfigfileurlorpath--options)

Aditya-A-G:
will create a fresh pr

PR #286, opened 2024-11-04
fix: preserve i18n.json formatting to avoid unnecessary PRs

Aditya-A-G:
fixes: #197 

Improved JSON serialization to preserve existing formatting in i18n.json. If a Prettier config is provided, it will be used for formatting; otherwise, the .editorconfig will be used, or defaults will be applied if neither is available.


Aditya-A-G (commented):
@maxprilutskiy @partik03 I think we should make getConfig function asynchronous. This would allow us to update the line to `await saveConfig(result);` and replace synchronous filesystem calls with asynchronous ones.

Additionally, these changes will necessitate updating any calls to the `getConfig` function to use await. Please let me know if you would like me to proceed with these modifications.

packages/cli/src/utils/config.ts
```diff
@@ -23,10 +24,19 @@ export function getConfig(resave = true): I18nConfig | null {
   return result;
 }
 
-export function saveConfig(config: I18nConfig) {
+export async function saveConfig(config: I18nConfig) {
   const configFilePath = _getConfigFilePath();
 
-  const serialized = JSON.stringify(config, null, 2);
+  const prettierConfigFilePath = await prettier.resolveConfigFile(configFilePath) || ''
+
```
maxprilutskiy: this is wrong, configFilePath is the location of the i18n.json file, please see line #47.

maxprilutskiy: also, prettier must find its config file itself: it can be in multiple formats, also it can be specified as a field in package.json, so let's let it do its job

packages/cli/src/utils/config.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  const configPath = path.join(process.cwd(), "i18n.json");
+
+  afterEach(() => {
+    if (fs.existsSync(prettierConfigPath)) {
+      fs.unlinkSync(prettierConfigPath);
+    }
+    if (fs.existsSync(configPath)) {
+      fs.unlinkSync(configPath);
+    }
+  });
```
maxprilutskiy, line 19: 1. Please mock filesystem during tests. Unit tests shouldn't mess with surrounding environment.
2. Add two test cases:
   - Test if i18n.json is saved in default format when no explicit prettier config exists
   - Test if i18n.json is saved using the format from explicit prettier config when one exists

maxprilutskiy (changes requested):
Thanks for the pull request, you're almost there! Left some comments, please take a look.

packages/cli/src/utils/config.ts
Aditya-A-G: Thank you for pointing that out! I see now that specifying configFilePath isn’t necessary since it points to i18n.json, and Prettier can locate its config file starting from process.cwd() on its own. I’ll remove this and let Prettier handle the config detection, as you suggested.

packages/cli/src/utils/config.spec.ts
Aditya-A-G: Thanks for the guidance! I’ll update the tests to mock the filesystem.

packages/cli/src/utils/config.ts
```diff
@@ -23,10 +24,19 @@ export function getConfig(resave = true): I18nConfig | null {
...
+  const prettierConfigFilePath = await prettier.resolveConfigFile() || ''
+
+  const prettierOptions = await prettier.resolveConfig(prettierConfigFilePath, {editorconfig: true}) || {
+    parser: 'json',
+    useTabs: false,
+    tabWidth: 2,
+  }
+    
+  const serialized = await prettier.format(JSON.stringify(config), prettierOptions);
+
```
maxprilutskiy, line 39: tbh this looks really good. how do we know it's working just looking at the code? be sure to add tests that showcase the core logic actually works here, and let's merge. good stuff!

maxprilutskiy:
@Aditya-A-G lmk if you need help :)

Aditya-A-G:
Hey @maxprilutskiy, Thanks for asking. I'm unavailable for the next few days due to my exams. If anyone wants to do the remaining meanwhile, then feel free to do so.

maxprilutskiy:
closing due to inactivity (lmk in discord if we should reopen it)

PR #301, opened 2024-11-07
fix(cli): add early exit for --frozen flag if lockfile update required

### LINKED ISSUE #110: Move `--frozen` flag check to a separate pre-localization step
**Description**  
Move the `--frozen` flag check to a separate step before starting the translation loop. This ensures the process fails fast if necessary, saving API calls and improving the developer experience.

**Proposed Solution**  
Refactor the translation process to move the `--frozen` flag checks before entering the main translation loop.

**References**
https://github.com/replexica/replexica/blob/main/packages/cli/src/i18n.ts

partik03:
Ensures that the --frozen flag exits before entering the localization loop if lockfile changes are detected, preventing unnecessary processing and API calls. Useful for CI/CD workflows to check localization changes.

Closes #110

### DIFF
--- a/.changeset/fuzzy-feet-buy.md
+++ b/.changeset/fuzzy-feet-buy.md
@@ -1,0 +1,5 @@
+---
+"@replexica/cli": minor
+---
+
+--frozen flag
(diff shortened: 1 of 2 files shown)


partik03:
@maxprilutskiy please review this one, whenever you'll get some time out

PR #404, opened 2025-01-24
feat(cli): interactive init command

mathio:
Init command accepts parameters:
- `-h, --help` show help
- `-f --force` overwrite existing config
- `-s --source <locale>` source locale
- `-t --targets <locale...>` list of target locales
- `-b, --bucket <type>` type of bucket
- `-p, --paths <path...>` list of paths for the bucket
- `-y, --no-interactive` Disable interactive mode

The command runs in interactive mode by default, asking each parameter.

After init completes:
- in interactive mode it asks user to:
    - open documentation
    - login to CLI, if not authenticated already
- in non-interactive mode it:
    - displays if user is authenticated
    - displays link to docs


.changeset/sharp-mangos-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@replexica/cli": minor
```
maxprilutskiy, line 2:
```suggestion
"@replexica/cli": minor
"replexica": minor
```

maxprilutskiy (approved):

PR #430, opened 2025-01-28
fix: gitlab & bitbucket actions

mathio:


action/src/platforms/_base.ts
```diff
@@ -27,17 +29,17 @@ export abstract class PlatformKit<PlatformConfig extends BasePlatformConfig = Ba
...
+      LINGODOTDEV_WORKING_DIRECTORY: Z.string().optional(),
     }).parse(process.env);
 
     return {
       replexicaApiKey: env.LINGODOTDEV_API_KEY,
       isPullRequestMode: env.LINGODOTDEV_PULL_REQUEST,
-      commitMessage: env.LINGODOTDEV_COMMIT_MESSAGE,
-      pullRequestTitle: env.LINGODOTDEV_PULL_REQUEST_TITLE,
-      workingDir: env.LINGODOTDEV_WORKING_DIRECTORY,
+      commitMessage: env.LINGODOTDEV_COMMIT_MESSAGE || defaultMessage,
```
mathio, line 40: In Github, when the input is omitted, the variable is still defined (with empty string value) and the `.default()` on `zod` above does not work. It works in other platforms, as they do not define the env variable when omitted (there are no "inputs", but they define env vars directly).

Using `||` here makes sure the default value is used when empty string is supplied, since both commit message and PR title can not be empty.

maxprilutskiy (approved):

PR #580, opened 2025-03-25
chore(packages/spec): add precise types for locales.ts

### LINKED ISSUE #560: Spec Module (/packages/spec):  Add more specific return types to locale utility functions
Some functions in `locales.ts` could benefit from more precise TypeScript return types. For example:

```
// Current implementation
export const getLocaleCodeDelimiter = (locale: string): string | null => {
 // ...
}
```

```
// Could be improved to
export const getLocaleCodeDelimiter = (locale: string): "-" | "_" | null => {
 // ...
}
```

TheRohit:
fixes #560


packages/spec/src/locales.ts
```diff
@@ -262,6 +263,6 @@ export const resolveOverridenLocale = (locale: string, delimiter?: "-" | "_" | n
   return locale.replace(currentDelimiter, delimiter);
 };
 
-export function normalizeLocale(locale: string) {
+export function normalizeLocale(locale: string): string {
```
mathio, line 266: This is not necessary since Typescript automatically infers simple types like this. But it also does not hurt to be explicit since the method is exported from the lib 🙃

mathio (approved):

mathio:
Please resolve conflicts and we can merge this!

Issue #569, opened 2025-03-25
SDK Modules (/packages/sdk):  Optimize the `countWordsInRecord` method for better performance


7hacker:
The current implementation of `countWordsInRecord` could be optimized for performance, especially when dealing with large payloads. Please also add unit tests for this optimization

```
private countWordsInRecord(payload: any | Record<string, any> | Array<any>): number {
 // Fast path for empty or null values
 if (!payload) return 0;
  // Use a more efficient algorithm for large arrays
 if (Array.isArray(payload)) {
   if (payload.length === 0) return 0;
   if (payload.length > 100) {
     // Use reduce for better performance with large arrays
     return payload.reduce((acc, item) => acc + this.countWordsInRecord(item), 0);
   }
   // Use loop for small arrays to avoid function call overhead
   let count = 0;
   for (let i = 0; i < payload.length; i++) {
     count += this.countWordsInRecord(payload[i]);
   }
   return count;
 }
  // Rest of the implementation...
}
```

Ashura707:
I would Like to take up this issue

7hacker:
@Ashura707 please submit a PR

Ashura707:
@7hacker made the pr.Could you check and let me know

mathio:
I find this a micro-optimization at best.  Unless we save more than few seconds, I'd argue that splitting code like this `if (payload.length > 100) {` leads to harder readable and maintainable code since it has 2 implementations of the same logic.

7hacker:
@mathio we can close it as a non issue if thats the case

mathio:
Let's close this one. 

@7hacker @Ashura707 sorry for not being on top of this before we opened a ticket.

PR #585, opened 2025-03-26
perf: optimize countWords function for large payloads

Ashura707:
Fix:#569
**Before:**
Screenshot 2025-03-26 100720

**After:**
Screenshot 2025-03-26 100547

->Removed recursion based approach and made it iterative to process it.Added test cases,
->Optimized processing speed of large payloads increasing the performance


Ashura707:
@maxprilutskiy could the lingo dev team check this and let me know

.changeset/four-cooks-sit.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
mathio, line 2: This needs a changeset if we want to release the change.

packages/sdk/src/index.ts
```diff
@@ -174,18 +174,43 @@ export class LingoDotDevEngine {
...
-    } else if (typeof payload === "object" && payload !== null) {
-      return Object.values(payload).reduce((acc: number, item) => acc + this.countWordsInRecord(item), 0);
-    } else if (typeof payload === "string") {
-      return payload.trim().split(/\s+/).filter(Boolean).length;
-    } else {
-      return 0;
+    const stack: any[] = [payload];
+    let totalWordCount = 0;
+  
+    while (stack.length > 0) {
```
mathio: How is this an optimization? If I am reading this right, the code is executed the same number of times as before. This is a micro-optimization at best (if `while` is faster than recursion), however I find this new code harder to read than before.

I'd like a second pair of eyes on this, @maxprilutskiy @7hacker please have a look.

mathio (changes requested):

packages/sdk/src/index.ts
Ashura707: @mathio I can come up with a better optimization for this making it more readable.Should I make a new pr or add the changes in this branch?

mathio: I raised my concerns in the issue. Lets continue the conversation there.

Ashura707: **New Optimizations**
Screenshot 2025-03-27 084005
->Improved execution time for large and deeply nested payloads.
->Lower memory consumption by eliminating redundant allocations.
->Removed Recursion because it might cause stack overflow.

mathio:
@Ashura707 we decided to take a different direction here. Good job optimizing the performance, however we prefer simplicity and better maintainability over maximum performance here.

Sorry for not raising it internally before we opened an issue.

PR #581, opened 2025-03-25
refactor(action): reduce redundancy in gitConfig for GitHub, GitLab and BitBucket

giteshsarvaiya:
Solves Issue: #573   
### Description:
**[Refactor]** This PR refactors the `gitConfig` implementation in the `./action/src` directory for GitHub, GitLab, and Bitbucket, reducing code redundancy.
It also introduces error handling using `try-catch` blocks for improved reliability.

PR #590, opened 2025-03-27
refactor: move gitConfig logic to parent and call it in child using super.gitConfig()

SXsid:
#581 move gitConfig logic to parent and execute in child using super.gitConfig()


maxprilutskiy:
fantastic job @SXsid !

Issue #592, opened 2025-03-27
`lingo.dev i18n --frozen` is broken with .po files


aroman:
here is my patch:

```patch
diff --git a/build/cli.cjs b/build/cli.cjs
index 509fb60a7d2f1ebe8da90ea0b4f600fd940d0d4c..d285bffc160103134b0560db533f3a201b679478 100644
--- a/build/cli.cjs
+++ b/build/cli.cjs
@@ -2756,6 +2756,9 @@ function variableExtractLoader(params) {
     pull: async (locale, input2) => {
       const result = {};
       for (const [key, value] of Object.entries(input2)) {
+        if (typeof value !== "string") {
+          continue;
+        }
         const matches = value.match(specifierPattern) || [];
         result[key] = result[key] || {
           value,
diff --git a/build/cli.mjs b/build/cli.mjs
index 67aab4ed1acfa5df600987061813571a2a07c99f..2fda5f840dc77f5b99e1182aee0879a2d30e1e51 100644
--- a/build/cli.mjs
+++ b/build/cli.mjs
@@ -2756,6 +2756,9 @@ function variableExtractLoader(params) {
     pull: async (locale, input2) => {
       const result = {};
       for (const [key, value] of Object.entries(input2)) {
+        if (typeof value !== "string") {
+          continue;
+        }
         const matches = value.match(specifierPattern) || [];
         result[key] = result[key] || {
           value,
```

mathio:
@aroman I can look at the root cause of this issue. 

Could you please share the `.po` file that is causing this? Thanks!

aroman:
@mathio this will happen on _any_ `.po` file as far as I can tell.

I just reprod with this blank .po file:

```po
msgid ""
msgstr ""
"POT-Creation-Date: 2025-02-04 21:01-0800\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=utf-8\n"
"Content-Transfer-Encoding: 8bit\n"
"X-Generator: @lingui/cli\n"
"Language: en\n"
"Project-Id-Version: \n"
"Report-Msgid-Bugs-To: \n"
"PO-Revision-Date: \n"
"Last-Translator: \n"
"Language-Team: \n"
"Plural-Forms: \n"
```

![Image](https://github.com/user-attachments/assets/d1238cab-2e82-4807-a761-f7a00d0d355a)

PR #593, opened 2025-03-27
fix: value.match is not a function with .po files

aroman:
this is just a quick and dirty workaround. to fix i18n --frozen not working with .po files.  see https://github.com/lingodotdev/lingo.dev/issues/592


mathio:
Thanks 🙌  I will test this locally and review later today.

mathio (approved):
Hello @aroman looks good 🚀 

In order for PR checks to pass and release this change please run `pnpm new` in repo root to add a new [changeset](https://github.com/changesets/changesets). I tried adding it for you, but I dont have write permissions to your fork.

Sorry I missed this one last week.

aroman:
@mathio thanks for the reply. I dont think this PR should be merged as-is - it was just a quick hack as I mentioned to illustrate the problem. Probably a bit deeper digging into why the types are not as expected is warranted here.

PR #622, opened 2025-04-10
feat: key renames

maxprilutskiy:


packages/cli/src/cli/utils/delta.ts
```diff
@@ -0,0 +1,100 @@
...
+      targetData: Record<string, any>;
+      checksums: Record<string, string>;
+    }) {
+      let added = _.difference(Object.keys(params.sourceData), Object.keys(params.targetData));
+      let removed = _.difference(Object.keys(params.targetData), Object.keys(params.sourceData));
+      const updated = _.filter(Object.keys(params.sourceData), (key) => {
+        return MD5(params.sourceData[key]) !== params.checksums[key] && params.checksums[key];
+      });
+
+      // Find renamed keys - keys that exist in both but with different names (same content)
```
mathio: Keys do not exist in both, value hashes do, right?

```suggestion
      // Find renamed keys - hashes that exist in both (same content) but with different keys
```

packages/cli/src/cli/cmd/i18n.ts
```diff
@@ -130,12 +133,73 @@ export default new Command()
...
+            await deltaProcessor.saveChecksums(checksums);
           }
         }
-        ora.succeed("i18n.lock created");
+        ora.succeed("Localization cache initialized");
       } else {
-        ora.succeed("i18n.lock loaded");
+        ora.succeed("Localization cache loaded");
+      }
+      // Handle json key renames
```
mathio, line 146: This is automatic on running `i18n`, right?

I can see 2 potential issues:

1. What if this is not actually a rename, but rather the same string was removed and added in different place? Eg. it could be translated differently because it is in different context.
2. I see the MD5 hash is calculated for all source file keys (to calculate `updated` array in delta processor), this might be slow for projects with 100s of files.

Shall we put this behind a flag (eg. `--migrate`) to run only when explicitly required?

mathio (approved):

PR #874, opened 2025-06-07
feat(cli/csv): improve `csv` loader

### LINKED ISSUE #870: Change / Support custom id/key column naming for `csv`
Currently, `csv` loader expects a mandatory lowercase `id` column which will be the main key, it would be great if you could declare it to something else, or automatically detect it.

And/or, it would be great if the default naming is changed to something like `KEY`, so it is less confusing, lowercase `id` is unintuitive and is conflicted / could be mis-understood for Indian/Indonesian, when the CSV file is using the columns as locales structuring like the official demo: https://github.com/lingodotdev/lingo.dev/blob/main/packages/cli/demo/csv/file.csv

NamesMT:
This PR is a set of changes for `csv` loader:
+ Reduce duplicated parsing
+ Dynamic key column naming support
  * Resolves #870

#### Implementation:

I added some variables like `inputParsed` to prevent parsing twice in both `pull` and `push`

I added a local function `_detectKeyColumnName()` to automatically detect the key column naming, current logics including: `known preferred keys` (`KEY`|`key`), `first non-empty cell`.

#### Additional Context:

At first I was working on another solution where I deprecate `id` (and still supports it for backward-compat), and prefers `KEY`, and adding a whole new option to set the `keyColumn`, but they requires too big of a change, I then followed @maxprilutskiy's suggestion of looking at `[0]`, its simpler and works too I guess, could be breaking for those who is using the old `id` naming in a non-first column, but intuitively there should be mostly none affected, safe for non-major?

### DIFF
--- a/packages/cli/src/cli/loaders/csv.ts
+++ b/packages/cli/src/cli/loaders/csv.ts
@@ -33,7 +33,7 @@ export default function createCsvLoader(): ILoader<
         if (detectedKeyColumnName) return detectedKeyColumnName;
 
         for (const key of ["KEY", "key"]) {
-          if (inputParsed.some((row) => row[key])) {
+          if (inputParsed[0][key]) {
             // console.debug(`Detected key column name from preferred keys: ${key}`);
             return keyColumnName = detectedKeyColumnName = key;
           }
--- a/packages/cli/src/cli/loaders/csv.ts
+++ b/packages/cli/src/cli/loaders/csv.ts
@@ -8,18 +8,47 @@ export default function createCsvLoader(): ILoader<
   string,
   Record<string, string>
 > {
+  let inputParsed: Record<string, any>[];
+  let keyColumnName: string = "KEY";
+  let detectedKeyColumnName: string | null = null;
+
   return createLoader({
-    async pull(locale, _input) {
-      const input = parse(_input, {
+    async pull(locale, input) {
+      inputParsed = parse(input, {
         columns: true,
         skip_empty_lines: true,
         relax_column_count_less: true,
-      });
+      }) as Record<string, any>[];
 
       const result: Record<string, string> = {};
 
-      _.forEach(input, (row) => {
-        const key = row.id;
+      /**
+       * Tries to detect the key column name
+       *
+       * Current detect logics and orders:
+       * + known preferred keys
+       * + first non-empty cell
+       */
+      function _detectKeyColumnName() {
+        if (detectedKeyColumnName) return detectedKeyColumnName;
+
+        for (const key of ["KEY", "key"]) {
+          if (inputParsed.some((row) => row[key])) {
+            // console.debug(`Detected key column name from preferred keys: ${key}`);
+            return keyColumnName = detectedKeyColumnName = key;
+          }
+        }
+
+        const firstContentfulRow = parse(input.split('\n').find(l => l.length)!)[0];
+        const firstContentfulColumn = firstContentfulRow.find(Boolean)
+        if (firstContentfulColumn) {
+          // console.debug(`Detected key column name from first non-empty cell: ${firstContentfulColumn}`);
+          return keyColumnName = detectedKeyColumnName = firstContentfulColumn;
+        }
+      }
+
+      _.forEach(inputParsed, (row) => {
+        const key = row[_detectKeyColumnName()];
         if (key && row[locale] && row[locale].trim() !== "") {
           result[key] = row[locale];
         }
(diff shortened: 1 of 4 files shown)


NamesMT:
Note: I intentionally did not add `id` as known preferred key because it would be bad in the future, as `id` is conflicted with `Indonesian` locale code.

packages/cli/src/cli/loaders/csv.ts
```diff
@@ -8,47 +8,70 @@ export default function createCsvLoader(): ILoader<
   string,
   Record<string, string>
 > {
+  let inputParsed: Record<string, any>[];
```
maxprilutskiy: you don't need inputParsed here. instead, use one of the parameters .push accepts, here's for the reference: https://github.com/lingodotdev/lingo.dev/blob/main/packages/cli/src/cli/loaders/_utils.ts#L97

packages/cli/src/cli/loaders/csv.ts
```diff
@@ -8,47 +8,70 @@ export default function createCsvLoader(): ILoader<
...
+  let keyColumnName: string = "KEY";
```
maxprilutskiy: let's not have "preferred" key name. Instead, treat the very first column as key column. you can fallback to `"KEY"` only if the name of the first column is empty.

that's also very important for backward compatibility.

(make sure it's empty after trimming, so we treat whitespace-only names as "empty" too)

packages/cli/src/cli/loaders/csv.ts
```diff
@@ -8,47 +8,70 @@ export default function createCsvLoader(): ILoader<
...
+  let detectedKeyColumnName: string | null = null;
```
maxprilutskiy: pls do not use variables in this scope (outside of pull/push), that will cause issues when we compose several loaders together. check https://github.com/lingodotdev/lingo.dev/blob/main/packages/cli/src/cli/loaders/_utils.ts#L97 - `.push` might already have the parameters you need.

packages/cli/src/cli/loaders/csv.ts
```diff
@@ -8,47 +8,70 @@ export default function createCsvLoader(): ILoader<
...
-      _.forEach(input, (row) => {
-        const key = row.id;
+      /**
+       * Tries to detect the key column name
+       *
+       * Current detect logics and orders:
+       * + known preferred keys
+       * + first non-empty cell
+       */
+      function _detectKeyColumnName() {
```
maxprilutskiy: please declare the function outside the loader - that way we can export it, and have surgical unit tests for it.

packages/cli/src/cli/loaders/csv.ts
```diff
@@ -8,47 +8,70 @@ export default function createCsvLoader(): ILoader<
...
+        const firstContentfulRow = parse(input.split('\n').find(l => l.length)!)[0];
+        const firstContentfulColumn = firstContentfulRow.find(Boolean)
+        if (firstContentfulColumn) {
+          // console.debug(`Detected key column name from first non-empty cell: ${firstContentfulColumn}`);
+          return keyColumnName = detectedKeyColumnName = firstContentfulColumn;
+        }
+      }
+
+      _.forEach(inputParsed, (row) => {
+        const key = row[_detectKeyColumnName()];
```
maxprilutskiy: calling function once, outside of the `forEach` loop is more efficient, since the value doesn't change

packages/cli/src/cli/loaders/index.spec.ts
```diff
@@ -104,7 +104,25 @@ describe("bucket loaders", () => {
   });
 
   describe("csv bucket loader", () => {
-    it("should load csv data", async () => {
+    it("should load csv data ('KEY' as key, preferred, second column)", async () => {
```
maxprilutskiy: update the tests too, the first column is always the key column

packages/cli/src/cli/loaders/csv.ts
NamesMT: Hi Max, I have tried earlier to return the parsed input from `pull` so that `push` can access it, but it bloat the lockfile with too many info, any ideas?

NamesMT:
Hi @maxprilutskiy , I've corrected most of your changes requests, however, I can't find a clean way to solve the remaining ones, which requires not using variables in the outer scope.
Using parameters like `data` / `pullOutput` like the link you referenced means that I have to return it via the result in `pull`, but doing so will bloat the lockfile.

packages/cli/src/cli/loaders/csv.ts
maxprilutskiy: replied on discord, let's discuss options

maxprilutskiy:
merged! 💯

PR #876, opened 2025-06-08
fix: timestamp update on Windows

vrcprl:


packages/compiler/src/lib/lcp/index.ts
```diff
@@ -148,8 +148,23 @@ export class LCP {
...
+            "⚠️  Lingo: Auto-reload disabled - system blocks Node.js timestamp updates",
+          );
+          console.warn(
+            "   💡 Fix: Adjust security settings to allow Node.js file modifications",
+          );
+          console.warn(
+            "   ⚡ Workaround: Manually refresh browser after translation changes",
+          );
+        }
+      }
```
maxprilutskiy: use dedent and chalk, show one console.log, and add recommendation to get help on discord

Issue #878, opened 2025-06-08
Missing GROQ_API_KEY with Next.js


bar-lavie:
### Description:

I've added to Groq API key to my env. file (used GROQ_API_KEY)

My `next.config.ts` looks like:

```ts
// ... next config

const withLingo = lingoCompiler.next({
  sourceLocale: "en",
  targetLocales: ["es", "fr", "de"],
  models: {
    "*:*": "groq:mistral-saba-24b",
  },
});

export default withSentryConfig(withMDX(withLingo(nextConfig)), {
// ... some sentry config.
```

When using: `npm run dev`, I get the following output in my terminal:
`💡 You're using Lingo.dev Localization Compiler in your project, which requires a GROQ API key to work.`
Which prevent the dev process to run

### Reference:
https://lingo.dev/en/compiler/frameworks/nextjs

## Update:
I'm using `vercel-env-pull` to fetch my env. variables from the vercel platform
The variables are being saved in `env.local` (default by vercel - as the file is being ignored by git) and not in the actual `.env` file 

Seems that lingo is looking for the variable only in the `.env` file which make sense, but I wonder why it's not taking in consideration `.env.local` as well as the next compiler includes it when running `next dev`

I'm not sure it's a bus/issue - but I believe that for a better DX it will be better to allow / include any env. file that are part of the next dev/build proceess

mathio:
Hi @bar-lavie thanks for mentioning this! We are adding support for your `.env.local` file too to improve the DX when using Next.js: #900

bar-lavie:
Thank you for the quick turnaround and sublime DX (: @mathio 
Working now in `"lingo.dev": "^0.97.1"`

PR #900, opened 2025-06-10
fix(compiler): load API key from env var and env files

mathio:


packages/compiler/src/utils/llm-api-key.ts
```diff
@@ -1,12 +1,21 @@
...
 
 // Generic function to retrieve key from process.env, with .env file as fallback
-function getKeyFromEnv(envVarName: string): string | undefined {
-  const ephemeralEnv = {} as Record<string, string>;
-  dotenv.config({ processEnv: ephemeralEnv });
-  return ephemeralEnv[envVarName];
+export function getKeyFromEnv(envVarName: string): string | undefined {
+  if (process.env[envVarName]) {
+    return process.env[envVarName];
+  }
```
mathio, line 10: Prioritize env var, per existing function definition in the comment above.

packages/compiler/src/utils/llm-api-key.ts
```diff
@@ -1,12 +1,21 @@
...
+  const result = dotenv.config({
+    path: [
+      path.resolve(process.cwd(), ".env"),
+      path.resolve(process.cwd(), ".env.local"),
+      path.resolve(process.cwd(), ".env.development"),
```
mathio, line 15: Resolves #878

PR #911, opened 2025-06-11
fix(compiler): hyphens in locale names

maxprilutskiy:


packages/compiler/src/utils/create-locale-import-map.ts
```diff
@@ -0,0 +1,20 @@
+import * as t from "@babel/types";
+
+export function createLocaleImportMap(
+  allLocales: string[],
+  dictionaryPath: string,
+): t.ObjectExpression {
+  return t.objectExpression(
+    allLocales.map((locale) =>
+      t.objectProperty(
+        t.stringLiteral(locale),
```
maxprilutskiy, line 10: instead of t.identifier

cc @mathio

Issue #851, opened 2025-06-05
[Feature Request] Support for Open Source Models


productboy:
LLM options are In settings.ts:
```
llm: {
      openaiApiKey: env.OPENAI_API_KEY || systemFile.llm?.openaiApiKey,
      anthropicApiKey: env.ANTHROPIC_API_KEY || systemFile.llm?.anthropicApiKey,
      groqApiKey: env.GROQ_API_KEY || systemFile.llm?.groqApiKey,
    },
```

Let's add support for open source models; for example OpenRouter:
```
llm: {
      openrouterApiKey: env.OPENROUTER_API_KEY || systemFile.llm?.openrouterApiKey,
      anthropicApiKey: env.ANTHROPIC_API_KEY || systemFile.llm?.anthropicApiKey,
      groqApiKey: env.GROQ_API_KEY || systemFile.llm?.groqApiKey,
    },
``` 

Or, support Ollama via the ai-sdk:
```
import { Ollama } from "ollama-ai-provider";
```

maxprilutskiy:
let's add it @productboy, great idea!

any specific model you'd like to try next?

productboy:
This open source model would be a great start:
 qwen3:8b

The-Best-Codes:
@maxprilutskiy I could work on this soon if you want :)

maxprilutskiy:
Upd: added Google AI models in #859 #875

Next step - to add Qwen

The-Best-Codes:
@productboy A PR was just merged that will add support for Ollama and another was merged that will add support for OpenRouter. When the updated packages are published, would your issue be solved?

mathio:
Compiler supports multiple LLM providers and models now. You can find all supported providers in thew docs:
https://lingo.dev/en/compiler/how-it-works#llm-providers

If you'd like support for more providers feel free to open a pull request or request specific provider in a new issue. 

**⚠️ Please request new providers one per issue, to keep issues actionable for contributors.**

productboy:
> [@productboy](https://github.com/productboy) A PR was just merged that will add support for Ollama and another was merged that will add support for OpenRouter. When the updated packages are published, would your issue be solved?

Yes!  Appreciate the quick turnaround.

PR #913, opened 2025-06-11
feat(compiler): ollama ai provider support

The-Best-Codes:
Addresses #851.

### DIFF
--- a/packages/cli/src/cli/localizer/explicit.ts
+++ b/packages/cli/src/cli/localizer/explicit.ts
@@ -57,7 +57,7 @@ export default function createExplicitLocalizer(
         factory: (_params) => createOllama().languageModel(provider.model),
         id: provider.id,
         prompt: provider.prompt,
-        authRequired: false,
+        skipAuth: false,
       });
   }
 }
@@ -68,9 +68,9 @@ function createAiSdkLocalizer(params: {
   prompt: string;
   apiKeyName?: string;
   baseUrl?: string;
-  authRequired?: boolean;
+  skipAuth?: boolean;
 }): ILocalizer {
-  const authRequired = params.authRequired !== false;
+  const authRequired = params.skipAuth !== false;
 
   const apiKey = process.env[params?.apiKeyName ?? ""];
   if ((authRequired && !apiKey) || !params.apiKeyName) {
--- a/packages/cli/src/cli/localizer/explicit.ts
+++ b/packages/cli/src/cli/localizer/explicit.ts
@@ -8,6 +8,7 @@ import { ILocalizer, LocalizerData } from "./_types";
 import { LanguageModel, Message, generateText } from "ai";
 import { colors } from "../constants";
 import { jsonrepair } from "jsonrepair";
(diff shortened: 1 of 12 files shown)


packages/cli/src/cli/localizer/explicit.ts
```diff
@@ -51,18 +52,28 @@ export default function createExplicitLocalizer(
...
 
 function createAiSdkLocalizer(params: {
-  factory: (params: { apiKey: string; baseUrl?: string }) => LanguageModel;
+  factory: (params: { apiKey?: string; baseUrl?: string }) => LanguageModel;
   id: NonNullable<I18nConfig["provider"]>["id"];
   prompt: string;
-  apiKeyName: string;
+  apiKeyName?: string;
   baseUrl?: string;
+  authRequired?: boolean;
```
maxprilutskiy: let's invert and do `doNotAuth?: boolean` instead: it's more conventional, and it'd be harded to make mistakes for someone who's not familiar with the codebase v much

The-Best-Codes:
There's an issue. You have to use colons (`:`) in Ollama model names to specify variants, but the code is splitting by provider:model and only taking the first chunk after the colon.
So for example, `ollama:qwen:0.6b` is parsed as `ollama` and `qwen`, when it should be `ollama` and `qwen:0.6b`.

The-Best-Codes:
Not yet ready to merge. Need to update error messages for providers that don't require auth.

The-Best-Codes:
Just found another bug or edge case where when using Ollama, after translating, it starts over again. I don't have the resources to test this efficiently, so I might need someone else's help.

The-Best-Codes:
Ready to merge, if you don't have any feedback!

packages/cli/src/cli/localizer/explicit.ts
```diff
@@ -51,35 +52,44 @@ export default function createExplicitLocalizer(
...
   id: NonNullable<I18nConfig["provider"]>["id"];
   prompt: string;
-  apiKeyName: string;
+  apiKeyName?: string;
   baseUrl?: string;
+  skipAuth?: boolean;
 }): ILocalizer {
-  const apiKey = process.env[params.apiKeyName];
-  if (!apiKey) {
+  const skipAuth = params.skipAuth == true;
```
The-Best-Codes: Should be `===` probably.

packages/compiler/src/lib/lcp/api/provider-details.ts
```diff
@@ -22,4 +22,11 @@ export const providerDetails: Record<
     getKeyLink: "https://ai.google.dev/",
     docsLink: "https://ai.google.dev/gemini-api/docs/troubleshooting",
   },
+  ollama: {
+    name: "Ollama",
+    apiKeyEnvVar: undefined, // Ollama doesn't use env vars for keys
```
The-Best-Codes: Need to fix the wording of these comments. Ollama doesn't require API keys period.

maxprilutskiy (approved):

PR #932, opened 2025-06-13
feat(compiler): support next.js with turbopack enabled

The-Best-Codes:
Working on compiler support for Turbopack in Next.js.

----

Edit:

## How to test this PR

1. Checkout the PR, reinstall deps, & rebuild:

   ```bash
   gh pr checkout 932
   pnpm install
   pnpm turbo build
   ```

2. Try a build in `demo/next-app`:

   ```bash
   # cd demo/next-app
   pnpm run build
   pnpm run start # try toggling translations in the browser
   ```

3. Enable Turbopack:

    You can optionally enable it in the Lingo config, or let it be auto-detected.

   ```diff
   // next.config.ts
   export default lingoCompiler.next({
     sourceLocale: "en",
     targetLocales: ["es", "ja", "fr", "ru", "de", "zh", "ar", "ko"],
     models: {
       "*:*": "groq:mistral-saba-24b",
     },
     useDirective: true,
   + turbopack: {
   +   enabled: true,
   + },
   })(nextConfig);
   ```

   You **must** at least make this change:

   ```diff
   -"build": "NODE_ENV=production next build",
   +"build": "NODE_ENV=production next build --turbo",
   ```

4. Try step `#2` again. It should work with Turbopack on!

----

For older versions of Next.js where the Turbopack config is in `experimental.turbo` instead of `turbopack`, you can set the `useLegacyTurbo` parameter in the Lingo.dev Turbopack object to `true`.


The-Best-Codes:
Not ready to merge yet FYI.

The-Best-Codes (commented):
Basically ready to merge, but I would like to go over my changes with a maintainer before merging anything. :)

mathio:
Hello @The-Best-Codes 

thank you for your PR. I just tested it locally and it works flawlessly! 

I see the logic from `load` and `transform` methods in [compiler/src/index.ts](https://github.com/The-Best-Codes/lingo.dev/blob/best/exp-turbopack-next/packages/compiler/src/index.ts) is duplicated in `lingoLoaderPath` in the [compiler/src/lingo-turbopack-loader.ts](https://github.com/The-Best-Codes/lingo.dev/blob/best/exp-turbopack-next/packages/compiler/src/lingo-turbopack-loader.ts). 

Can you extract this logic into single place, please? This will help in future so we dont need to update it multiple places. 

Thanks!

The-Best-Codes:
@mathio I centralized most of it into `_loader-utils.ts`, does it look okay?

mathio (approved):
This looks good, thanks 👍 

Unless @maxprilutskiy has any objections we will merge this tomorrow (on Friday) 🚀

mathio:
I just learned there is a `process.env.TURBOPACK` env variable available when running with turbopack. Do you think we could use that to detect it instead of introducing new compiler config? Eg:

```js
const isTurbopackEnabled = process.env.TURBOPACK === '1' || process.env.TURBOPACK === 'true';
```

@The-Best-Codes what do you think? Or is there any specific reason why this detection is not reliable and you'd prefer to have an explicit config?

Thanks!

The-Best-Codes:
@mathio 

Yep! I found out about that and was thinking about doing it in a separate PR, but I can do it in this one! You can also detect if the `--turbo` or `--turbopack` flag is present.

So my thinking was basically this. I will update the config to default to "auto" instead of false. So basically, `turbopack.enabled` could be `true | false | "auto"`. I will get to work on that right away!

mathio:
Do we need the setting at all? Why not leave it automatic and detect via
env var? I think no setting is simpler, unless you know about a case when
we want to force it to be true/false.

I tried using "argv" to detect the --turbo flag too, but it was not present
in every process on "next dev".

The-Best-Codes:
@mathio I think it's best to leave a setting especially for two main things:

- The user may want to force Turbo on or off
- The user may be using an older Next.js where `turbopack` is `experimental.turbo`, in which case they need to change that.

mathio:
Ok, that makes sense 👍

The-Best-Codes:
@mathio Updated!

maxprilutskiy:
@The-Best-Codes 👏 👏 👏

PR #955, opened 2025-06-23
feat(cli): progressively push translated chunks to locale files

### LINKED ISSUE #886: CLI: Cache translated chunks as they are received
I was trying to generate locales for my project, which has ~1,400 entries in a single `en.json` file for three different locales. However, I reached the limit with just one CLI command, and all progress on already translated entries, which was displayed in the CLI (one of them was nearly 80%), has been completely lost. It would be better to update files based on a response, rather than waiting for localization to finish.

mathio:
# 🆕 Progressive push / caching

Push received chunks progressively to target locale files. For bucket types storing all locales in single file update single file as the chunks are received.

This ensures the translations are processed as received. This works when running with `--force` flag as well. If the process fails, the process will not re-start from the beginning (unless restarted with `--force` flag). Instead of `run --force`I suggest to use the new `purge` command (see below) in combination with `run` command.

Instead of caching chunks to separate location target locale files are populated directly (resolves #886).

**Note:** This functionality is only available for localizers that support progress callbacks. At the moment this is only *Lingo.dev* localizer, other LLM localizers do not support this functionality.

### Allow pushing partial results into bucket loaders

Bucket loaders can accept partial data. They do not expect complete data in single push.

### Refactor key-order functionality

Do not merge source locale keys into target locale file.

To preserve keys order use separate `createEnsureKeyOrderLoader` that orders keys in target locale based on source locale order.

### Refactor some loaders

Do not merge source locale keys into target locale file.

Instead remember which keys were removed and add back only specific keys.

### Still push final translated result at the end

At the end of `run` command still push the final object. This ensures compatibility with localizers without progress callback (all BYOK localizers at the moment).

# 🆕 New `purge` command

Removes translations for given --bucket, --file, --key, --locale
Asks for interactive confirmation for each file unless --yes-really is
present.

# 🧹 Code cleanup: Remove "isCacheRestore" option

Not used anymore


packages/cli/src/cli/cmd/run/execute.ts
```diff
@@ -191,40 +190,58 @@ function createWorkerTask(args: {
...
                   assignedTask,
                   percentage: progress,
                 });
               },
             );
 
-            let finalTargetData = _.merge(
+            const finalTargetData = _.merge(
               {},
               sourceData,
```
mathio, line 237: Do not include `sourceData`?

packages/cli/src/cli/index.ts
```diff
@@ -59,6 +58,7 @@ Star the the repo :) https://github.com/LingoDotDev/lingo.dev
   .addCommand(statusCmd)
   .addCommand(mayTheFourthCmd, { hidden: true })
   .addCommand(run, { hidden: true })
```
mathio:
```suggestion
  .addCommand(run)
```

packages/cli/src/cli/loaders/ensure-key-order.ts
```diff
@@ -0,0 +1,50 @@
...
+      }
+      dataKeys.delete(key);
+    }
+  }
+
+  // should not happen unless there is LLM halucination? 👇
+  // Add keys that are in `data` but not in `original` at the end
+  // for (const key of dataKeys) {
+  //   orderedData[key] = data[key];
+  // }
```
mathio:
```suggestion
```

packages/cli/src/cli/localizer/lingodotdev.ts
```diff
@@ -26,6 +26,8 @@ export default function createLingoDotDevLocalizer(
   const engine = new LingoDotDevEngine({
     apiKey: auth.apiKey,
     apiUrl: auth.apiUrl,
+    batchSize: 1,
+    idealBatchItemSize: 1,
```
mathio:
```suggestion
```

PR #1007, opened 2025-07-05
feat: support comments in i18n.json

davidturnbull:
The goal of this PR is to allow comments in the `i18n.json` file by parsing them as JSONC: https://jsonc.org/

This has a couple of benefits:

- if users have complex configurations, they can adequately comment them
- we can include comments in the code blocks that contain configuration (while still allowing users to copy/paste them)


mathio:
@davidturnbull love this idea 😍  I was missing this without even realizing it!

PR #1010, opened 2025-07-07
feat: improve type-safety of CompilerParams

davidturnbull:
This PR attempts to improve the type-safety of the `CompilerParams` type:

- set `sourceLocale` to `LocaleCode` instead of `string`
- set `targetLocale` to `LocaleCode` instead of `string`
- require the locale pairings in the `models` property to match one of the following formats:
  - `locale:locale`
  - `*:locale`
  - `locale:*`
  - `*:*`
- require the model identifier to be a colon separated string (e.g., `string:string`)

The model identifier can definitely be typed better — I'm imagining constructing a type of exact providers and models — but that's a bigger task.

## Example

image

### DIFF
--- a/packages/compiler/src/_base.ts
+++ b/packages/compiler/src/_base.ts
@@ -101,8 +101,11 @@ export type ModelMap = {
 /**
  * A pairing of a source and target locale.
  */
-export type SourceTargetLocale = LocalePair | AnyTargetLocale | AnySourceLocale | AnyLocale;
-
+export type SourceTargetLocale =
+  | LocalePair
+  | AnyTargetLocale
+  | AnySourceLocale
+  | AnyLocale;
 
 /**
  * A translation from a specific source locale to a specific target locale.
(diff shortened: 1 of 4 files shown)


packages/compiler/src/_base.ts
```diff
@@ -89,6 +90,44 @@ export type CompilerParams = {
...
+ * A mapping between locale pairings and the model to use to translate that pairing.
+ */
+export type ModelMap = {
+  [key in SourceTargetLocale]?: ModelIdentifier;
+};
+
+/**
+ * A pairing of a source and target locale.
+ */
+export type SourceTargetLocale = AnyTargetLocale | AnySourceLocale | AnyLocale;
```
mathio: ModelMap can define not just wildcard locales (`en:*`, `*:en`, `*:*`) but also specific locale pairs (`en:es`, `fr:en`, etc.). If I read the code right, this one is missing.

```suggestion
/**
 * A pairing of a source and target locale.
 */
export type SourceTargetLocale = LocalePair | AnyTargetLocale | AnySourceLocale | AnyLocale;


/**
 * A translation from a specific source locale to a specific target locale.
 */
export type LocalePair = `${LocaleCode}:${LocaleCode}`;
```

davidturnbull: Whoops. Not sure how I missed the most obvious one. 😅

PR #1015, opened 2025-07-08
feat: create script to generate CLI reference docs

davidturnbull:
This PR adds a script that generates CLI reference docs based on the docs within the CLI itself. The goal is to be able to automatically update reference docs on the website every time a new version of the CLI is released.

For the time being, the script must be run manually — there is no automation that moves the file to the documentation website. But this provides the foundations that should make that possible.

See `scripts/docs/README.md` for usage instructions.

This comment has the example output: https://github.com/lingodotdev/lingo.dev/pull/1015#issuecomment-3055042185

**Note:** My understanding is that, when using the `pull_request_target` event, the changes to the action will only run after it exists on `main`. To test the action, I've been testing with the `pull_request` event.


mathio:
@davidturnbull your understanding of GHA events is correct 👍

mathio:
The script only produces a markdown file at the moment, correct? There is no integration with website (docs) repo, right?

davidturnbull:
> The script only produces a markdown file at the moment, correct? There is no integration with website (docs) repo, right?

@mathio Correct. For the time being, I'll copy things over manually. But hope to address the full automation within the next couple of weeks. 😄

mathio:
Awesome! 💪

PR #1040, opened 2025-07-11
fix(compiler): handle type-only react imports in compiler

The-Best-Codes:
## Problem

The compiler was failing when encountering type-only React imports like `import type React from "react"`. When the compiler needed to add runtime imports (e.g., `Fragment`, `LingoComponent`, etc.), it would incorrectly try to add them to the existing type-only import declaration, causing runtime errors.

**Example fail:**

```tsx
import type React from "react";

export function Component() {
  return <>Hello world</>;  // ReferenceError: Fragment is not defined
}
```

The compiler would incorrectly transform this to:

```tsx
import type React, { Fragment } from "react";  // Fragment not available at runtime
```

## Solution

Fixed the `getOrCreateImport` utility function to properly handle type-only imports:

1. **`findExistingImport`**: Skip type-only imports when looking for existing runtime imports
2. **`createImportDeclaration`**: Only add new imports to existing non-type import declarations

This fix should apply to all compiler features that need runtime imports, like Fragment transformations, `LingoComponent`, etc.

**Previous fail, now fixed:**

```tsx
import type React from "react";
import { Fragment } from "react";  // Separate runtime import

export function Component() {
  return <Fragment>Hello world</Fragment>;  // Works correctly
}
```

I also added a new test, "should handle type-only React import and add Fragment import".

### DIFF
--- a/.changeset/modern-hairs-poke.md
+++ b/.changeset/modern-hairs-poke.md
@@ -1,5 +1,5 @@
 ---
-"@lingo.dev/_compiler": minor
+"@lingo.dev/_compiler": patch
 ---
 
 Fixed the compiler to handle type-only react imports.


The-Best-Codes:
If you want an easy way to test this, go to `demo/next-app` and change `src/components/hero-subtitle.tsx`:

```diff
"use server";
"use i18n";
+import type React from "react";

export async function HeroSubtitle() {
  return (
+  <>
      <p className="text-lg sm:text-xl text-gray-600 mb-10 max-w-xl mx-auto leading-relaxed">
        Localize your React app in every language in minutes.
        <br />
        Scale to millions from day one.
      </p>
+  </>
  );
}
```

If you are on the main branch (without my changes applied), this will cause an error when you run build or dev. If you checkout my PR & rebuild it should work!

packages/compiler/src/jsx-fragment.spec.ts
```diff
@@ -106,6 +106,26 @@ function Component() {
...
+import type React from "react";
+import { Fragment } from "react";
+function Component() {
+  return <Fragment></Fragment>;
+}
+`.trim();
+    const result = runMutation(input);
+    expect(result).toBe(expected);
+  });
+
```
maxprilutskiy, line 128: this is some 12/10 work @The-Best-Codes, amazing! 👏 👏 👏

.changeset/modern-hairs-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@lingo.dev/_compiler": minor
```
maxprilutskiy:
```suggestion
"@lingo.dev/_compiler": patch
```

maxprilutskiy (approved):

maxprilutskiy:
Well done, @The-Best-Codes! 💯

PR #1049, opened 2025-07-16
feat(sdk): added new sdk methods localizeStringArray

### LINKED ISSUE #568: SDK Modules (/packages/sdk): Implement convenience methods for common use cases
The SDK could benefit from some convenience methods that encapsulate common usage patterns, such as:

```
// Localize a simple key-value map
async localizeSimpleMap(
 map: Record<string, string>,
 params: Z.infer<typeof localizationParamsSchema>
): Promise<Record<string, string>> {
 return this.localizeObject(map, params);
}
```

```
// Localize an array of strings
async localizeStringArray(
 strings: string[],
 params: Z.infer<typeof localizationParamsSchema>
): Promise<string[]> {
 const mapped = strings.reduce((acc, str, i) => {
   acc[`item_${i}`] = str;
   return acc;
 }, {} as Record<string, string>);
  const result = await this.localizeObject(mapped, params);
  return Object.values(result);
}
```

VAIBHAVSING:
fixes #568 
### New SDK Features:
* **`localizeStringArray` Method**: Added to localize arrays of strings while preserving their order. The method maps the array to an object for localization and then converts it back to an array.

### Testing Enhancements:
* **Tests for `localizeStringArray`**: Added unit tests to ensure arrays are localized correctly and maintain their order, including handling of empty arrays.


packages/sdk/src/index.ts
```diff
@@ -298,6 +298,47 @@ export class LingoDotDevEngine {
...
+   *   - targetLocale: The target language code (e.g., 'es')
+   *   - fast: Optional boolean to enable fast mode (faster for bigger batches)
+   * @returns A new map with the same keys but localized string values
+   */
+  async localizeSimpleMap(
+    map: Record<string, string>,
+    params: Z.infer<typeof localizationParamsSchema>,
+  ): Promise<Record<string, string>> {
+    return this.localizeObject(map, params);
+  }
```
maxprilutskiy: Actually, this is not needed, since it has 0 extra logic. Having just localizeObject is fine. The original github issues seemes to be a bit misleading (that's on us). Please remove `localizeSimpleMap` and the related test.

packages/sdk/src/index.ts
```diff
@@ -298,6 +298,47 @@ export class LingoDotDevEngine {
...
+        acc[`item_${i}`] = str;
+        return acc;
+      },
+      {} as Record<string, string>,
+    );
+
+    const result = await this.localizeObject(mapped, params);
+    return Object.values(result);
+  }
+
```
maxprilutskiy, line 325: looks good!

VAIBHAVSING:
PTAL @maxprilutskiy

maxprilutskiy (approved):

Issue #1079, opened 2025-07-29
Make CLI play music


maxprilutskiy:
Add music playback functionality to the CLI.

## Implementation Details

Special mode of `run` command that enables sound effects when `--sound` flag is provided.

## Sound Effect Triggers:

- **Subtask completed**: Light chime or bell sound
- **Task section completed**: Upbeat completion sound (like a level-up)
- **Entire command completes successfully**: Triumphant tada/fanfare music  
- **Command fails**: Sad/disappointed sound effect (like a deflating balloon)

## Technical Requirements:

- Cross-platform compatibility
- Brief and non-intrusive sound effects
- Consider using system sounds or embedding small audio files
- Make it fun and engaging without being annoying
- Allow users to customize or disable specific sounds
- **Top quality, clean code that is easy to read and maintain**
- **Surgical updates only** - if any changes are needed in existing run command files, make minimal, precise modifications

## Creative Direction:

Open-ended task - focus on making CLI usage more enjoyable and rewarding through well-timed, contextually appropriate sound feedback that enhances the user experience.

ankur0904:
@maxprilutskiy 
I am working on this issue. Soon, raise the PR.

maxprilutskiy:
resolved in #1157

PR #1157, opened 2025-09-06
feat(cli): add --sound flag for task completion

ankur0904:
After all main task will completed it make the `success` sound
 
 **Main Task**
- setup()
- plan()
- execute()
- renderSummary()

If any of the task failed or the `run` command will fail it'll generate the `failure` sound.

> I have taken the music file from [here](https://pixabay.com/sound-effects/)

#1079


packages/cli/src/cli/cmd/run/index.ts
```diff
@@ -69,7 +103,14 @@ export default new Command()
     "Debounce delay in milliseconds for watch mode (default: 5000ms)",
     (val: string) => parseInt(val),
   )
+  .option(
+    "--sound",
+    "Play sound on completion, partially completion and failed of the task"
+  )
   .action(async (args) => {
+    // log the args in terminal
```
The-Best-Codes: Is there a specific reason you added this, or was it just for debugging during development? If you're done with development and it wasn't there before, maybe remove it :)

packages/cli/assets/failure.mp3
The-Best-Codes, line 1: This is probably waaay too picky of me:
Compared to the success sound effect, this sound effect is really quiet. I think it would be nicer if I could hear it a little better.

The-Best-Codes (changes requested):
Looking good so far! There are a couple of major things that are left out, but I'll put those in a separate comment so you'll be sure to see them 😀

The-Best-Codes:
In my review, I mentioned a big issue. Here are the details:

The sound assets **won't be published to npm**, since `files` is explicitly set in the `package.json`.
https://github.com/lingodotdev/lingo.dev/blob/764d1f0b22f548eac2ace6d850e29362aab269fe/packages/cli/package.json#L99-L102
image

To fix this, add `assets` to the files list in the `package.json`.

ankur0904:
@The-Best-Codes 
Hi, thank you for reviewing the PR. I have addressed all the changes you mentioned in your review. :)

vrcprl (approved):
Looks great! I really enjoyed testing it.

ankur0904:
@vrcprl @The-Best-Codes 
Thank you so much for merging the PR. I really enjoyed working in this issue.

PR #1249, opened 2025-10-27
feat: add metadata registry for buckets

davidturnbull:
## Problem

Bucket types (json, yaml, markdown, etc.) have different capabilities, but there was no programmatic way to query these capabilities. Questions like "which buckets support the `[locale]` placeholder?" or "which buckets support locked keys?" required manually reviewing loader implementations.

This made it difficult to:
- Generate accurate documentation
- Validate user configurations before instantiating loaders
- Build UI components that need to display bucket capabilities
- Understand bucket features without deep-diving into code

## Solution

Added a metadata registry that explicitly declares capabilities for all 29 bucket types.

### New `Bucket` Type

Each bucket now includes:
- `displayName` - Human-readable label (e.g., "JSON", "Xcode XCStrings")
- `supportsFormatter` - Can use code formatters (Prettier, etc.)
- `supportsInjectLocale` - Can inject locale code into content
- `supportsLockedKeys` - Supports locking specific translation keys
- `supportsIgnoredKeys` - Supports ignoring specific keys
- `supportsLockedPatterns` - Supports pattern-based locking (MDX only)
- `supportsLocalePlaceholder` - Supports `[locale]` placeholder in file paths
- `createLoader()` - Factory function to create the loader

### Architecture

Moved from a large switch statement to a **standalone registry approach**:

```typescript
export const BUCKETS: Record<string, Bucket> = {
  json: {
    displayName: "JSON",
    supportsFormatter: true,
    supportsLocalePlaceholder: true,
    // ... other capabilities
    createLoader: (ctx) => composeLoaders(...)
  },
  // ... all 29 bucket types
}
```

This approach was chosen because:
1. **Lightweight consumption** - Metadata can be imported without loader dependencies
2. **Build-time usage** - Documentation generation doesn't need full loader infrastructure
3. **Clear separation** - Metadata (declarative) is separate from loaders (runtime behavior)
4. **Queryability** - Easy to lookup by bucket type without function calls

### Public API

- `BUCKETS` - Complete registry of all bucket types
- `getBucket(type)` - Helper to lookup a bucket by type
- `Bucket` type - For consumers who need the type definition

## Impact

- No breaking changes - `createBucketLoader()` signature unchanged
- Enables future features like auto-generated documentation
- Makes bucket capabilities explicit and maintainable
- Simplifies validation and error messaging

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/cli/src/cli/loaders/index.ts
+++ b/packages/cli/src/cli/loaders/index.ts
@@ -66,17 +66,21 @@ export type BucketLoaderContext = {
 };
 
 export type Bucket = {
+  displayName: string;
   supportsFormatter: boolean;
   supportsInjectLocale: boolean;
   supportsLockedKeys: boolean;
   supportsIgnoredKeys: boolean;
   supportsLockedPatterns: boolean;
   supportsLocalePlaceholder: boolean;
-  createLoader: (ctx: BucketLoaderContext) => ILoader<void, Record<string, any>>;
+  createLoader: (
+    ctx: BucketLoaderContext,
+  ) => ILoader<void, Record<string, any>>;
 };
 
 export const BUCKETS: Record<string, Bucket> = {
   android: {
+    displayName: "Android XML",
     supportsFormatter: false,
     supportsInjectLocale: false,
     supportsLockedKeys: true,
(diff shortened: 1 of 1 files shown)


maxprilutskiy:
moving to the new arch

Issue #1206, opened 2025-10-05
Request: Support for dynamic text translation in React


iam-avijit-me:
Hey team,

I would like to request a feature to support **dynamic text translation** in React components. 

For example, currently we can translate static keys like this:

locales/en.json
```
{
  "error1": "Welcome, dear!",
  "error2": "Hello, friend!"
}
```

In a React component:
```
import { Translate } from 'lingo.dev/react/client';

<Alert>{Translate('error1')}</Alert>
```

It would be great if we could also **easily translate dynamic text**, such as messages generated in switch statements or functions, without hardcoding each message.

Thanks for your amazing work!

codesofsandeep:
Hi everyone 👋,

I noticed that @SaketTiwari2008 has already added a commit related to this feature request (support for dynamic text translation in React).  
I wanted to check whether this implementation is currently under review or if the issue is still open for contributions.

If it's not yet finalized, I'd be happy to help by testing, reviewing, or improving the implementation.  
Please let me know how I can assist!

Thanks for maintaining such a great project 🙌

sumitsaurabh927:
how is it different from Lingo.dev [compiler](https://lingo.dev/en/compiler)?

maxprilutskiy:
@iam-avijit-me what's the DX you're going after here? mind showing an example of how the final result would look like, potentially?

maxprilutskiy:
This is being solved by Lingo.dev Compiler, and is out of scope for Lingo.dev CLI.

Reference:
- https://github.com/lingodotdev/lingo.dev/tree/main/packages/new-compiler
- https://lingo.dev/compiler

PR #1262, opened 2025-10-29
added support for dynamic text translation in React components (Fixes…

SaketTiwari2008:
**Description** :--

This PR adds support for **dynamic text translation** in React components.
Currently, lingo.dev only supports translating predefined static keys from locale files.
With this update, developers can now translate runtime-generated or dynamic strings without hardcoding each message.

---
**Changes** :--

- Enhanced LingoComponent to handle dynamic text values.
- Added a fallback helper that attempts to translate runtime strings.
- Maintains full backward compatibility with existing static key translations.

---
**Testing** :--

Verified that both static and dynamically generated messages render correctly and fallback gracefully when translations are missing.

---
**Notes** :--

- No breaking changes introduced
- No new dependencies added
- Fully compatible with existing locale JSON structure.

---
**Fixes**: #1206
Type:  Feature Addition
Title: Support for Dynamic Text Translation in React Components.


maxprilutskiy:
Closing as:

1. the suggested code isn't solving the issue. 
2. Dynamic values are already supported.

PR #1279, opened 2025-10-29
issue resolved successfully

sahilkapase:
I located the cache logic and updated it to use plain JSON for reading/writing, added a legacy reader that avoids executing code, and switched the filename to a JSON extension across the compiler.
What I changed

Switched dictionary filename to JSON
Updated LCP_DICTIONARY_FILE_NAME from dictionary.js to dictionary.json.
Rewrote cache read/write to use JSON and avoid Function
ensureDictionaryFile now creates {} instead of export default {};.

_read:
Reads dictionary.json as JSON with JSON.parse.
If the new file doesn’t exist, detects a legacy dictionary.js, strips the export default wrapper, and parses the inner JSON safely with JSON.parse (no code execution). Returns the parsed cache.

_write:
Writes pure JSON (no export default) and formats it with Prettier as JSON.
Prettier formatting for JSON
_format now prefers parser: "json" by default.

Kept loader behavior stable
All places using LCP_DICTIONARY_FILE_NAME (e.g., loadInclude, RSC loader, path utilities) now refer to dictionary.json, so dynamic imports and compilation still work as before.

Tests updated and added
Updated existing LCPCache tests to expect JSON content instead of export default {...};.
Added migration test to ensure:
If only legacy dictionary.js exists with export default { ... };, reading works without executing code.
On first write, a new dictionary.json is written with pure JSON, ensuring a seamless migration.

Notes
Migration happens on first write: legacy content is read safely and the next write stores the JSON file. No code execution is used at any stage.
Deterministic formatting and key ordering are preserved.
No changes are needed in user code; imports continue to work since the plugin intercepts dictionary.json requests.
I updated the constant to dictionary.json, changed cache read/write to JSON (no Function), added legacy fallback read, and adjusted tests to validate JSON behavior and migration.


sumitsaurabh927:
@sahilkapase you need to update the PR title and desc as per the guidelines in our contributor's guide.

PR #1318, opened 2025-10-29
fix: replace NaN% with 0.0% when all translations complete

lokyaan:
### Summary
Fixes a division-by-zero issue in the `lingo.dev status` CLI output.
When `totalWordsToTranslate` equals 0, `(0 / 0) * 100` evaluates to `NaN`, causing the CLI to print:

### Fix
Added a guard to return `0.0` instead of `NaN` when total words = 0:
```ts
const percent =
  totalWordsToTranslate === 0
    ? 0.0
    : ((words / totalWordsToTranslate) * 100).toFixed(1);


###Result
- fr: ~0 words (0.0% of total)

###Impact
Prevents CI/CD regex validation failures.
Improves CLI reliability and output consistency.

### DIFF
--- a/packages/cli/src/cli/cmd/status.ts
+++ b/packages/cli/src/cli/cmd/status.ts
@@ -515,7 +515,10 @@ export default new Command()
         console.log(`• Per-language breakdown:`);
         for (const locale of targetLocales) {
           const words = totalWordCount.get(locale) || 0;
-          const percent = totalWordsToTranslate===0?0.0:((words / totalWordsToTranslate) * 100).toFixed(1);
+          const percent =
+            totalWordsToTranslate === 0
+              ? 0.0
+              : ((words / totalWordsToTranslate) * 100).toFixed(1);
           console.log(
             `  - ${locale}: ~${words.toLocaleString()} words (${percent}% of total)`,
           );

### CODE REVIEW

File: package-lock.json
Comment (The-Best-Codes) [inline package-lock.json:1]: This file needs to be removed, the repo uses `pnpm`

File: package.json [hunk 1]
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@changesets/changelog-github": "^0.5.0",
     "@changesets/cli": "^2.27.10",
+    "build": "^0.1.4",
```
The-Best-Codes: Why is this change needed?

The-Best-Codes (changes requested):
Just a couple of things I see that you might want to change.
Also, not sure if you ran `pnpm run format` before committing but don't forget to do that 🙂

lokyaan:
Hi!!! Glad to see the review
So These are the following changes I made after this email
1. Removed package.json
2. The "build": "^0.1.4" got added due to dependency issue and this is
currently not present in the code
And right now I made sure to run   pnpm run format before committing
Thank Youuuu

lokyaan:
Hiiiii Lingo.dev
Any update regarding this????

The-Best-Codes:
@lokyaan I want to test this locally to see if the build dependency is actually needed on my end. AFAIK it works just fine without it.

lokyaan:
Yess it works when i tested it again after removing the dependency

The-Best-Codes:
@lokyaan Okay, maybe remove the dependency, then :)

lokyaan:
Hiii @lingodotdev/lingo.dev
***@***.***>
I thought I already did it and realised I was looking into another
package.json

I found out what u were referring to and removed the build dependency and
run it again and pushed the code

Thank You

The-Best-Codes:
@lokyaan Don't forget to run `pnpm install` to update the lockfile and `pnpm run format` to format the codebase again. After that I think this PR is ready to merge :grinning:

lokyaan:
After running these two commands, I got a lot of files that needs to be
committed. Should I add, commit and push all those to my branch???
I'm sry that this is taking time but I'm a bit confused at this point if
I'm dng that right

lokyaan:
Awesome, thanks!

PR #1330, opened 2025-10-29
docs: improved contribution guide for new contributors

Nischay23:


CONTRIBUTING.md
```diff
@@ -33,7 +33,11 @@ cd lingo.dev
 pnpm install
 ```
 
-Next, configure an AI API key. You can use either a Lingo.dev API key or one of the supported LLM model providers.
+Next, configure an AI API key. You can use either a Lingo.dev API key or one of the supported LLM model providers. A simple way to do this is to create a `.env` file in the root of the project and add your key, like this:
```
The-Best-Codes: Not sure that this clarification is very necessary, as the docs and information about the API key discovery process are right below. I guess it might be useful for new contributors, though.
One thing you might do while you're at it is update the documentation link right below, which is outdated, to `https://lingo.dev/en/cli/quick-start#step-3-configure-an-llm-provider`.

Nischay23: Thanks for the feedback!
I updated the outdated link and refined the wording as suggested.
Please let me know if you'd like any more adjustments

PR #1356, opened 2025-10-30
fix: improve error handling and documentation in locale regex and fetch response

Nitin-kumar-yadav:
The `export const LOCALE_REGEX` statement is defining a regular expression object named
`LOCALE_REGEX` that is used for parsing locale strings. The regular expression is constructed using
the `RegExp` constructor with the following components:

This code snippet is checking if the `metadataUrl` variable is falsy (in this case, `undefined`).
  If `metadataUrl` is `undefined`, it means that the `import.meta.url` feature is not supported or
  not available in the current environment. In that case, the code throws an error with the message
  "import.meta.url is undefined" to indicate that the script relies on `import.meta.url` and cannot
  proceed without it.


sumitsaurabh927:
Hi @Nitin-kumar-yadav please resolve the merge conflicts

PR #1477, opened 2025-10-31
replace root README with content from packages/cli/README.md

naayann:
Replaced the placeholder README.md (which only pointed to packages/cli/README.md) with the actual content from packages/cli/README.md so that the root README properly explains the project.


maxprilutskiy:
This is by design. Next time please create an issue first. Thank you.

Issue #1343, opened 2025-10-30
docs: create a Lingo.dev CLI integration guide for Electron


davidturnbull:
## Context

We're building a library of integration guides that show developers how to use **Lingo.dev CLI** with popular frameworks and platforms. Each guide is intended to be a complete, hands-on tutorial that takes someone from zero to having a working multilingual application.

## Task

Create a **Lingo.dev CLI** integration guide for Electron

The guide should:

- Follow the same general flow, scope, and structure of existing guides
- Use [react-intl](https://formatjs.github.io/docs/getting-started/installation/) to render translations

## Example guides

These are some examples of existing guides:

- Godot: https://lingo.dev/en/cli/frameworks/godot
- Mintlify: https://lingo.dev/en/cli/frameworks/mintlify
- Laravel: https://lingo.dev/en/cli/frameworks/laravel
- iOS: https://lingo.dev/en/cli/frameworks/ios
- React Native (Expo): https://lingo.dev/en/cli/frameworks/react-native-expo

## Caveat: Merging PRs

Our documentation content lives in an internal repo. This means **if you open a pull request for an integration guide, we can't merge it directly**. Instead, we use the following workflow to ensure everyone receives credit for their work:

1. Create a pull request for the current repo.
2. In the PR, include a Markdown file that contains the integration guide (e.g., `docs/guide.md`).
2. We'll review the PR, working with you to get the guide into the best state possible.
3. If we publish the guide on the website, we'll include a byline in the documentation that links to your GitHub profile.
4. Once the documentation has been published, we'll close the PR with a comment that references the document.

## Testing

Be sure to test the complete guide by carefully following all of the steps. (This is the most time-consuming part.)

gagan052:
Hey @davidturnbull , Hope you are doing well. I wanted to contribute to this project so I thought of starting with a documentation contribution.

Can you assign it to me ?

ranjithreddykv:
Hey @davidturnbull, I am just getting started with open source contribution can you please assign me this issue , to start my journey in open source contribution

sumitsaurabh927:
Hey @gagan052 do you wanna work on this issue?

mannubaveja007:
can i work on this issue?

Athaxv:
Hi @davidturnbull , 
Please check out PR #1500, it includes the complete and tested setup for Electron + Lingo.dev CLI + React-Intl, with all fixes and documentation updates in place.

[...]

PR #1500, opened 2025-10-31
docs: add Electron integration guide for Lingo.dev CLI

Athaxv:
This PR adds a new integration guide for Electron application using Lingo.dev CLI and React-Intl.

The guide includes:
- Electron + esbuild setup
- Lingo.dev CLI authentication and .env configuration
- React-Intl-based translation switching
- npx and local CLI fallback handling
- Automated translation sync using lingo.dev run

Author: @Athaxv

### DIFF
--- a/docs/electron-integration-guide.md
+++ b/docs/electron-integration-guide.md
@@ -1,0 +1,377 @@
+# Electron
+
+This guide shows you how to integrate Lingo.dev CLI with an Electron application using React and `react-intl` for internationalization.
+
+## Prerequisites
+
+Before you begin, make sure you have:
+
+- Node.js (v16 or higher) and npm installed
+- Basic knowledge of JavaScript, React, and Electron
+
+## Step 1: Create a New Electron Application
+
+Start by creating a new Electron project. Create a new directory and initialize it:
+
+```bash
+mkdir my-electron-app
+cd my-electron-app
+npm init -y
+```
+
+Install Electron as a dependency:
+
+```bash
+npm install electron
+```
+
+Create a basic Electron main process file `main.js`:
+
+```javascript
+const { app, BrowserWindow } = require('electron')
+const path = require('path')
+
+function createWindow() {
+  const win = new BrowserWindow({
+    width: 800,
+    height: 600,
+    webPreferences: {
+      nodeIntegration: false,
+      contextIsolation: true
+    }
+  })
+
+  win.loadFile(path.join(__dirname, 'renderer/index.html'))
+}
+
+app.whenReady().then(createWindow)
+
+app.on('window-all-closed', () => {
+  if (process.platform !== 'darwin') {
+    app.quit()
+  }
+})
+```
+
+Update your `package.json` to include the main entry point:
+
+```json
+{
+  "main": "main.js",
+  "scripts": {
+    "start": "npm run build && electron ."
+  }
+}
+```
+
+## Step 2: Set Up React
+
+Install React and React DOM:
+
+```bash
+npm install react react-dom
+```
+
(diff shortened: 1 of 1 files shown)


docs/electron-integration-guide.md
```diff
@@ -0,0 +1,377 @@
...
+
+```json
+{
+  "app.greeting": "Bonjour!",
+  "app.changeLanguage": "Changer la langue"
+}
+```
+
+
+## Step 7: npm run start
```
davidturnbull, line 275: This step is empty.

docs/electron-integration-guide.md
```diff
@@ -0,0 +1,377 @@
...
+```bash
+npm install --save-dev lingo.dev
+```
+
+**Important:** In Electron environments, the standard `npx` commands from the documentation may not work properly. Instead, use the direct Node.js commands.
+
+Initialize Lingo.dev in your project:
+
+```bash
+node ./node_modules/lingo.dev/bin/cli.mjs init
```
davidturnbull, line 200: This isn't how users should be calling the CLI.

They should be calling `npx lingo.dev@latest init`. It should work because `npx` isn't being used "in Electron".

docs/electron-integration-guide.md
```diff
@@ -0,0 +1,377 @@
...
+This guide shows you how to integrate Lingo.dev CLI with an Electron application using React and `react-intl` for internationalization.
+
+## Prerequisites
+
+Before you begin, make sure you have:
+
+- Node.js (v16 or higher) and npm installed
+- Basic knowledge of JavaScript, React, and Electron
+
+## Step 1: Create a New Electron Application
```
davidturnbull, line 12: Is there a tool that Electron provides for scaffolding a project, rather than these manual steps? (If not, that's fine.)

docs/electron-integration-guide.md
Athaxv: Yes you are right, but the npx command wasn’t running directly inside the Electron environment because Electron doesn’t inherit the full shell context or environment variables where npx is normally available. This can cause failures for users running commands through Electron rather than a regular terminal.
For users running it inside Electron or similar restricted environments, we can provide an alternative using a script in package.json:

"scripts": {
  "lingo": "node ./node_modules/lingo.dev/bin/cli.js"
}

This allows them to run:
npm run lingo init

So the docs can show both options, with npx as the standard method and the script-based command as a fallback for environments where npx isn’t directly accessible.

docs/electron-integration-guide.md
Athaxv: Sorry about that! I’ll fix it up right away and update the docs accordingly

docs/electron-integration-guide.md
Athaxv: Yes, there are tools like Electron Forge and Electron Builder, which many developers use as faster alternatives for scaffolding and packaging.

In this guide, I’ve kept the setup manual to clearly show how lingo.dev integrates with Electron and React-Intl, but we can certainly mention Forge or similar tools as optional faster alternatives if you think that would make the docs more useful.

I can update the docs accordingly if you’d like.

sumitsaurabh927:
@Athaxv please see if you can add @gagan052 as a co-author in light of [this](https://github.com/lingodotdev/lingo.dev/issues/1343#issuecomment-3482185335)

gagan052:
Thanks a lot, @sumitsaurabh927 Appreciate it 🙌
@Athaxv, great work on the PR — glad to be listed as a co-author!

sumitsaurabh927:
@Athaxv are you interested in continuing?

PR #1552, opened 2025-11-07
Update README language table

jimjoice2804:


maxprilutskiy:
Table is not needed, "•" is by design.

Closing.

PR #1555, opened 2025-11-08
feat: add Bengali localization support and update locales in LocaleSwitcher

pauldebanshu19:
…itcher


maxprilutskiy:
Closing, as part of the cleanup process.

1. PR doesn't reference any relevant reported issue;
2. The PR breaks the readme symlink;
3. Adds locale to the demo locale switcher, which doesn't make any difference.

PR #1598, opened 2025-11-12
fix(cli): uptake response_mime_type setting for gemini to avoid json parse issue

### LINKED ISSUE #1248: [Feature Request] Expose `response_mime_type` setting for Google Gemini provider to enable structured JSON output
Problem

  When using Google Gemini models (gemini-2.5-pro, gemini-2.5-flash) as a provider, translations intermittently fail with markdown-wrapped JSON responses instead of
  raw JSON:

  Unexpected token '`', "```json\n{\n..." is not valid JSON

  Root Cause

  Gemini models support structured JSON output via the response_mime_type: "application/json" parameter (https://ai.google.dev/gemini-api/docs/structured-output),
  but Lingo.dev's i18n.json schema does not expose this setting.

  Current schema supports:
  {
    "provider": {
      "id": "google",
      "model": "gemini-2.5-pro",
      "prompt": "...",
      "settings": {
        "temperature": 0  // ← Only setting available
      }
    }
  }

  Gemini needs:
  {
    "provider": {
      "id": "google",
      "model": "gemini-2.5-pro",
      "prompt": "...",
      "settings": {
        "temperature": 0,
        "response_mime_type": "application/json",  // ← Missing
        "response_schema": {}  // ← Optional but useful
      }
    }
  }

  Current Behavior

  - Small files (<8KB): Usually succeed with careful prompt engineering
  - Large files (>10KB): Frequently fail with markdown code blocks despite explicit prompt instructions

  Expected Behavior

  Setting response_mime_type: "application/json" forces Gemini to output raw JSON without markdown formatting, eliminating this class of errors entirely.

  Reproduction

  1. Create i18n.json with Gemini provider:
  {
    "provider": {
      "id": "google",
      "model": "gemini-2.5-pro",
      "prompt": "Translate from {source} to {target}. Output raw JSON only."
    }
  }

  2. Run translation on a large file (>15KB):
  npx lingo.dev@latest run --target-locale es --file "large-file.json"

  3. Observe intermittent failures with markdown-wrapped JSON

  Proposed Solution

  Extend i18n.json schema to support Gemini-specific settings:

  // i18n.json schema
  interface ProviderSettings {
    temperature?: number;

    // Gemini-specific
    response_mime_type?: "application/json" | "text/x.enum";
    response_schema?: object;  // JSON schema for output structure

    // OpenAI-specific (future)
    response_format?: { type: "json_object" };
  }

  Map these to Gemini's generationConfig:
  // When provider.id === "google"
  const config = {
    temperature: settings.temperature,
    responseMimeType: settings.response_mime_type,  // ← Add this
    responseSchema: settings.response_schema        // ← Add this
  };

  Workaround (Current)

  Heavy prompt engineering with temperature: 0:
  {
    "provider": {
      "prompt": "You are a JSON translation API. Your output MUST be parseable JSON. FORBIDDEN: Do NOT write ```json or ``` or any text outside the JSON object. 
  OUTPUT FORMAT: Start with { and end with }. No markdown, no code blocks...",
      "settings": { "temperature": 0 }
    }
  }

  This works ~70% of the time but is unreliable for large files.

kmr-rohit:
## Summary
Fixes #1248 

This change updates the translation response generation logic to use generateObject for JSON outputs. With this, the output will be a parsed JSON object instead of relying on markdown-wrapped or string-wrapped JSON.

What’s changed:

- The translation pipeline now calls generateObject(...) when the provider expects raw JSON output.
- Adjusted the schema/settings to pass the correct content type ("application/json") and ensure the object output is clean and free of markdown/code-block wrappers.


maxprilutskiy:
We found out that at scale, in our specific context, generateText works much better vs generate object and is more predictable. The discrepancies in format, if any, when using with google provider, can be solved by adjusting the message samples.

Additionally, ` if (params.id === "google" && params.settings?.responseMimeType) ... else ...` looks a bit odd, and is a code smell: typically if behaviour is different, it's a signal the strategy pattern should be applied.

Please propose your solution in #1248 first, get it approved, then open a pr. Thank you!

PR #1705, opened 2025-12-23
feat(compiler): move translation service initialization to see the logs

AleksandrSl:
## Summary
We don't really need to initialize it inside the server, since server is only a way to use the service in the dev mode.

Logs of the server are written to a file due to it being started from process which doesn't have access to console. But we want to see TranslationService initialization logs in the console, and the clearest way is to start it early, rather than extracting validation into a separate function.

## Changes

- Shows TranslationService logs in the console instead of log file
- Sets up e2e tests for compiler

## Checklist

- [x] Changeset added (if version bump needed)
- [x] No breaking changes (or documented below)


packages/new-compiler/src/plugin/next.ts
```diff
@@ -205,14 +206,12 @@ export async function withLingo(
     `Initializing Lingo.dev compiler. Is dev mode: ${isDev}. Is main runner: ${isMainRunner()}`,
   );
 
-  // TODO (AleksandrSl 12/12/2025): Add API keys validation too, so we can log it nicely.
```
AleksandrSl, line 208: That's what is solved in the PR (beside e2e tests and logs clenaup)

packages/new-compiler/src/translation-server/translation-server.ts
```diff
@@ -747,10 +711,6 @@ export class TranslationServer {
     try {
       const parsedLocale = parseLocaleOrThrow(locale);
 
-      if (!this.translationService) {
-        throw new Error("Translation service not initialized");
-      }
```
AleksandrSl, line 752: Quite dumb check, we make sure `TranslationService` is never null in constructor, since it's not something heavy  we should delay creating

packages/new-compiler/src/translators/pseudotranslator/index.ts
```diff
@@ -21,42 +21,20 @@ export class PseudoTranslator implements Translator<PseudoTranslatorConfig> {
   ) {}
 
   translate(locale: LocaleCode, entries: Record<string, TranslatableEntry>) {
-    this.logger.debug(
-      `[TRACE-PSEUDO] translate() ENTERED for ${locale} with ${Object.keys(entries).length} entries`,
-    );
     const delay = this.config?.delayMedian ?? 0;
     const actualDelay = this.getRandomDelay(delay);
 
-    this.logger.debug(
```
AleksandrSl, line 30: These were mostly added by LLM for debugging, we can add them back with LLM if needed

.github/workflows/pr-check.yml
```diff
@@ -71,3 +71,47 @@ jobs:
       - name: Require changeset to be present in PR
         if: github.event.pull_request.user.login != 'dependabot[bot]'
         run: pnpm changeset status --since origin/main
+
+  compiler-e2e:
```
AleksandrSl, line 75: Why this is a separate job:
- this test should probably be optional for now
- it's easier to see which tests failed exactly, e2e or the usual ones

It needs the previous check to pass because if usual tests fail, I don't think it makes a lot of sense to run e2e.

packages/new-compiler/package.json
```diff
@@ -123,11 +123,11 @@
     "clean": "rm -rf build",
     "test": "vitest --run",
     "test:watch": "vitest -w",
-    "test:prepare": "pnpm build && tsx tests/helpers/prepare-fixtures.ts",
-    "test:e2e": "playwright test --reporter=list",
-    "test:e2e:next": "playwright test --grep next --reporter=list",
-    "test:e2e:vite": "playwright test --grep vite --reporter=list",
-    "test:e2e:shared": "playwright test tests/e2e/shared --reporter=list",
+    "test:e2e:prepare": "tsx tests/helpers/prepare-fixtures.ts",
+    "test:e2e": "playwright test",
```
AleksandrSl, line 127: Reporter is set in the config

packages/new-compiler/package.json
```diff
@@ -123,11 +123,11 @@
     "clean": "rm -rf build",
     "test": "vitest --run",
     "test:watch": "vitest -w",
-    "test:prepare": "pnpm build && tsx tests/helpers/prepare-fixtures.ts",
-    "test:e2e": "playwright test --reporter=list",
-    "test:e2e:next": "playwright test --grep next --reporter=list",
-    "test:e2e:vite": "playwright test --grep vite --reporter=list",
-    "test:e2e:shared": "playwright test tests/e2e/shared --reporter=list",
+    "test:e2e:prepare": "tsx tests/helpers/prepare-fixtures.ts",
```
AleksandrSl, line 126: Build is required on turbo side

.github/workflows/pr-check.yml
```diff
@@ -71,3 +71,47 @@ jobs:
...
+        with:
+          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Install Playwright Browsers
```
AleksandrSl, line 109: Playwright says caching browsers isn't worth it - https://playwright.dev/docs/ci#caching-browsers (and we use only one)

packages/new-compiler/src/plugin/build-translator.ts
```diff
@@ -67,10 +64,6 @@ export async function processBuildTranslations(
 
   logger.info(`🌍 Build mode: ${buildMode}`);
 
-  if (metadataFilePath) {
```
AleksandrSl, line 70: This was internal info

PR #1955, opened 2026-02-03
feat(new-compiler): migrate metadata storage from lockfile to LMDB

### LINKED ISSUE #1636: Lingo.dev compiler fails with “TypeError: traverse is not a function” in React Router v7 + Vite
### What's happening?

First of all, thanks for the awesome project!

I’m running into an issue when using React Router v7+ with Vite.
When starting the development server, the Lingo.dev compiler fails during localization.

### Exact Reproduction Steps

1. Created an empty react router project
2. Config
3. Run

### Logs

ℹ️  Starting Lingo.dev compiler...

🔑  LLM API keys detected for configured providers: lingo.dev.
• Lingo.dev API key loaded from environment variables (LINGODOTDEV_API_KEY).

✨  ➜  Local:   http://localhost:5173/
    ➜  Network: http://192.168.1.68:5173/
    ➜  Network: http://10.125.0.14:5173/
    ➜  press h + enter to show help

⚠️  Lingo.dev compiler failed to localize your app
⚠️  Details: TypeError: traverse is not a function
    at _hasFileDirective (file:///Users/jarvis/Dev/dana/node_modules/@lingo.dev/_compiler/build/chunk-SRWLOKRJ.mjs:20393:3)
    at hasI18nDirective (file:///Users/jarvis/Dev/dana/node_modules/@lingo.dev/_compiler/build/chunk-SRWLOKRJ.mjs:20404:10)
    at file:///Users/jarvis/Dev/dana/node_modules/@lingo.dev/_compiler/build/chunk-SRWLOKRJ.mjs:20529:39
    at file:///Users/jarvis/Dev/dana/node_modules/@lingo.dev/_compiler/build/chunk-SRWLOKRJ.mjs:10:20
    at file:///Users/jarvis/Dev/dana/node_modules/@lingo.dev/_compiler/build/chunk-SRWLOKRJ.mjs:35:34
    at Function.thru (/Users/jarvis/Dev/dana/node_modules/lodash/lodash.js:8859:14)
    at /Users/jarvis/Dev/dana/node_modules/lodash/lodash.js:4430:28
    at arrayReduce (/Users/jarvis/Dev/dana/node_modules/lodash/lodash.js:697:21)
    at baseWrapperValue (/Users/jarvis/Dev/dana/node_modules/lodash/lodash.js:4429:14)
    at LodashWrapper.wrapperValue (/Users/jarvis/Dev/dana/node_modules/lodash/lodash.js:9114:14)

⚠️  Lingo.dev compiler failed to localize your app
⚠️  Details: TypeError: traverse is not a function

AndreyHirsa:
## Summary

Migrate metadata storage from JSON files with lockfile to LMDB database for improved concurrent access and elimination of ELOCKED errors. 

## Changes

- Replace proper-lockfile with LMDB: Use lmdb-js for metadata storage with MVCC (Multi-Version Concurrency Control) for safe concurrent access across bundler workers    
- Remove proper-lockfile dependency: No longer needed with LMDB's built-in concurrency handling
- Update related docs 

## Testing

- [x] Basic operations with db
- [x] Stats accuracy tracking
- [x] Cleanup logic
- [x] Error handling
- [x] All tests pass locally

## Visuals

N/A - No UI changes 

## Checklist

- [x] Changeset added
- [x] Tests cover business logic

Closes #1636

## Summary by CodeRabbit

* **New Features**
  * Metadata storage migrated from a JSON file to an LMDB-backed database with separate dev/build metadata directories and short-lived access.
  * Server runtime API updated to accept translation hashes and return { t, locale, translations }.

* **Documentation**
  * Compiler, architecture, transformation pipeline, and runtime docs updated to reflect new storage layout and APIs.

* **Tests**
  * Added comprehensive metadata tests covering creation, save/load, cleanup, large batches, and concurrency-like scenarios.

* **Chores**
  * Removed file-locking dependency in favor of LMDB-backed persistence.


packages/new-compiler/src/metadata/manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+} from "./manager";
+import type { TranslationEntry } from "../types";
+
+function createTestEntry(
+  overrides: Partial<TranslationEntry> & {
+    hash?: string;
+    sourceText?: string;
+  } = {},
+): TranslationEntry {
+  const hash = overrides.hash ?? `hash_${Math.random().toString(36).slice(2)}`;
```
AleksandrSl: nit: we probably can use a real hash function here, to avoid extra mocks?

packages/new-compiler/src/metadata/manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+  return {
+    type: "content",
+    hash,
+    sourceText: overrides.sourceText ?? `Source text for ${hash}`,
+    context: { filePath: "test.tsx", componentName: "TestComponent" },
+    location: { filePath: "test.tsx", line: 1, column: 1 },
+    ...overrides,
+  } as TranslationEntry;
```
AleksandrSl: Why does it not comply to the type so we have to cast?

packages/new-compiler/src/metadata/manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  describe("loadMetadata", () => {
+    it("should return empty metadata for new database", async () => {
+      const metadata = await loadMetadata(testDbPath);
+      expect(metadata.entries).toEqual({});
+      expect(metadata.stats!.totalEntries).toBe(0);
+    });
+
+    it("should load and preserve all entry fields", async () => {
+      const entry: TranslationEntry = {
```
AleksandrSl: nit: createTestEntry?

packages/new-compiler/src/metadata/manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+      // Update existing entry (count should not increase)
+      await saveMetadata(testDbPath, [
+        createTestEntry({ hash: "entry-1", sourceText: "v2" }),
+      ]);
+      const updated = await loadMetadata(testDbPath);
+      expect(updated.stats!.totalEntries).toBe(3);
+      expect(updated.entries["entry-1"].sourceText).toBe("v2");
+
+      // Empty array should not change anything
+      await saveMetadata(testDbPath, []);
```
AleksandrSl, line 102: nit: i think this tests is a bit to big in terms of responsibility, it tests several things (there is even and in the name 😄 )

packages/new-compiler/src/metadata/manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+      );
+
+      await saveMetadata(testDbPath, entries);
+      expect((await loadMetadata(testDbPath)).stats!.totalEntries).toBe(100);
+    });
+
+    it("should maintain data integrity after many operations", async () => {
+      // Many saves with overlapping keys
+      for (let i = 0; i < 10; i++) {
+        await saveMetadata(testDbPath, [
```
AleksandrSl: Let's make the saves here and above concurrent? They won't be fully parallele, but we may test at least asynchronous work.
I mean - run a lot of saves in the `Promise.all()`

AleksandrSl: Ah, lol, I see they are below, I think we can keep only concurrent versions then to keep amount of the tests reasonable, since concurrent adds more values, and there is nothing specific to catch in the synchronous test

packages/new-compiler/src/metadata/manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+      const final = await loadMetadata(testDbPath);
+      // 5 persistent + 10 unique = 15
+      expect(final.stats!.totalEntries).toBe(15);
+    });
+  });
+
+  describe("concurrent access (single process)", () => {
+    it("should handle concurrent operations from multiple calls", async () => {
+      // LMDB handles concurrent writes via OS-level locking
+      const promises = Array.from({ length: 10 }, async (_, i) => {
```
AleksandrSl: let's make it 100 or better 1000 if that doens't take much time?

packages/new-compiler/src/metadata/manager.test.ts
```diff
@@ -0,0 +1,219 @@
...
+        environment: "development",
+      });
+      expect(customResult).toContain(".custom-lingo");
+    });
+  });
+
+  describe("error handling", () => {
+    it("should throw descriptive error for invalid path", async () => {
+      const invalidPath = "/root/definitely/cannot/create/this/path";
+      await expect(loadMetadata(invalidPath)).rejects.toThrow();
```
AleksandrSl: I was going to say that we have tested that no error should be thrown 😄 but that was cleanup.

So we have to call createMetadata before trying to load it?
I wonder how this would work in concurrent scenario, should every caller call createEmptyMetadata?


Regarding the test, we do not test here that the error is descriptive, so we can drop this word from the name or/and add snapshot for the error text

packages/new-compiler/src/metadata/manager.ts
```diff
@@ -1,12 +1,78 @@
...
+const STATS_KEY = "__stats__";
+const METADATA_DIR_DEV = "metadata-dev";
+const METADATA_DIR_BUILD = "metadata-build";
+
+/**
+ * Opens an LMDB connection for a single operation.
+ *
+ * lmdb-js deduplicates open() calls to the same path (ref-counted at C++ level),
+ * so this is cheap. Each open() also clears stale readers from terminated workers.
+ * Closing before the loader returns ensures the handle is released even if the
```
AleksandrSl: Where exactly we do the closing here 🤔

packages/new-compiler/src/metadata/manager.ts
```diff
@@ -1,12 +1,78 @@
...
+ * worker is killed right after.
+ */
+function openDatabaseConnection(dbPath: string): RootDatabase {
+  fs.mkdirSync(dbPath, { recursive: true });
+
+  const isBuildMode = dbPath.endsWith(METADATA_DIR_BUILD);
```
AleksandrSl: I'd prefer to pass it as an argument, rather than infer a thing we now already at the call site.

packages/new-compiler/src/metadata/manager.ts
```diff
@@ -1,12 +1,78 @@
...
+
+/**
+ * Opens an LMDB connection for a single operation.
+ *
+ * lmdb-js deduplicates open() calls to the same path (ref-counted at C++ level),
+ * so this is cheap. Each open() also clears stale readers from terminated workers.
+ * Closing before the loader returns ensures the handle is released even if the
+ * worker is killed right after.
+ */
+function openDatabaseConnection(dbPath: string): RootDatabase {
```
AleksandrSl: As a small improvment if we mostly opening and closing the database in every method, not keeping the connection alive, we can create method like `runWithDatabaseConnect` or something similar which will do 

```
function runWithDbConnection(f: (db: Database) -> void) {
try {
  const db = open()
  f(db)
} finally {
  close()
}
}
```

So we simplify the work if anyone will be adding new methods, they won't forget to close the DB.

packages/new-compiler/src/metadata/manager.ts
```diff
@@ -17,174 +83,75 @@ export function createEmptyMetadata(): MetadataSchema {
...
   }
 }
 
 /**
- * Get the absolute path to the metadata file
+ * Persists translation entries to LMDB.
  *
- * @param config - Config with sourceRoot, lingoDir, and environment
- * @returns Absolute path to metadata file
+ * Uses transactionSync to batch all writes into a single commit. Async puts
```
AleksandrSl: I think we can decrease the level of details here and only leave the things that may affect the user. e.g. knowing that this function calls transactionSync means little

packages/new-compiler/src/metadata/manager.ts
```diff
@@ -17,174 +83,75 @@ export function createEmptyMetadata(): MetadataSchema {
...
-        );
-        return;
-      }
-      throw error;
-    } finally {
-      // Best-effort cleanup if rename failed for some reason
-      await fsPromises.unlink(tmpPath).catch(() => {});
-    }
+      const totalKeys = db.getKeysCount();
+      const entryCount =
```
AleksandrSl: We can check if we use entryCount anywhere and drop it entirely, I think it was added by an LLM and it was almost free to count it before, bow that we have to get stats and then do a `- 1`(why - by the way) it's time to consider the pros of this field

packages/new-compiler/src/plugin/transform/TRANSFORMATION_PIPELINE.md
```diff
@@ -155,38 +157,51 @@ export async function Welcome() {
...
-**Metadata Structure**:
+**Storage**: LMDB key-value database in `.lingo/metadata-dev/` or `.lingo/metadata-build/`
 
-```json
+**Why LMDB?**
+
+- ~1M ops/sec write speed vs ~50K for SQLite
+- Zero-copy reads from memory-mapped files
+- Built-in LZ4 compression
+- Simple key-value API without SQL overhead
```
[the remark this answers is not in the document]
AleksandrSl: Interesting 🤔  I do not have an opinion on LMDB vs Sqlite myself, we do not care about >100k ops per sec I believe, the easier the setup and the better it handles concurrent writes, the better

packages/new-compiler/src/plugin/unplugin.ts
```diff
@@ -381,7 +379,10 @@ export const lingoUnplugin = createUnplugin<
             trackEvent(TRACKING_EVENTS.BUILD_ERROR, {
               framework: currentFramework,
               errorType: "transform",
-              errorMessage: error instanceof Error ? error.message : "Unknown transform error",
+              errorMessage:
```
AleksandrSl, line 382: Interesting that it formats the code another way, did I forget to setup prettier 🤔

packages/new-compiler/src/metadata/manager.test.ts
AndreyHirsa: Switched to using the generateTranslationHash - no more mocks needed.

packages/new-compiler/src/metadata/manager.test.ts
AndreyHirsa: Right, that was ugly. Refactored.

packages/new-compiler/src/metadata/manager.test.ts
AndreyHirsa: Done.

packages/new-compiler/src/metadata/manager.test.ts
AndreyHirsa: Fair, split it into separate tests.

packages/new-compiler/src/metadata/manager.test.ts
AndreyHirsa: Makes sense. Dropped the sequential test and kept only the Promise.all version

packages/new-compiler/src/metadata/manager.test.ts
AndreyHirsa: Went with 1000, runs ~ 2 sec so no issues there.

packages/new-compiler/src/metadata/manager.test.ts
AndreyHirsa: loadMetadata and saveMetadata both go through runWithDbConnection -> openDatabaseConnection, which creates the directory automatically. In a concurrent scenario, mkdirSync({ recursive: true }) is safe - multiple callers hitting it simultaneously is fine

createEmptyMetadata was removed because it only existed to construct the old ` { entries: {}, stats: { totalEntries, lastUpdated } }` shape.

On the test - dropped "descriptive" from the name and added an assertion on the actual message (Failed to open LMDB at ${invalidPath}).

packages/new-compiler/src/metadata/manager.ts
AndreyHirsa: Refactored.

packages/new-compiler/src/metadata/manager.ts
AndreyHirsa: Refactored.

packages/new-compiler/src/metadata/manager.ts
AndreyHirsa: Good point. Refactored.

packages/new-compiler/src/metadata/manager.ts
AndreyHirsa: Refactored.

packages/new-compiler/src/metadata/manager.ts
AndreyHirsa: Yep, there is no real value in keeping this. Statistics have been removed. Metadata is now just a Record<string, TranslationEntry>

PR #1984, opened 2026-02-14
fix(cli): respect dev.usePseudotranslator config to skip API validation

### LINKED ISSUE #1983: fix(cli): respect dev.usePseudotranslator config to skip API key validation
## Bug Description

When `dev.usePseudotranslator: true` is set in `i18n.json`, the CLI still requires API keys to be configured. This should not happen - the pseudotranslator works entirely locally without any external APIs.

### Current Behavior
```json
// i18n.json
{
  "provider": "google",
  "dev": {
    "usePseudotranslator": true
  }
}
```

Running `lingo.dev run` still requires `GOOGLE_API_KEY` to be set, even though the pseudotranslator doesn't use it.

### Root Cause
In `packages/cli/src/cli/cmd/run/setup.ts`:
```typescript
const provider = ctx.flags.pseudo ? "pseudo" : ctx.config?.provider;
```

The CLI **only** checks for the `--pseudo` CLI flag. It does NOT check `ctx.config?.dev?.usePseudotranslator`. 

The config-based pseudotranslator setting is only evaluated inside `TranslationService`, which is created **after** the localizer setup and API key validation.

### Expected Behavior
When `dev.usePseudotranslator: true` is set in config, the CLI should:
1. Skip API key validation
2. Use the pseudo localizer (same as `--pseudo` flag)
3. Work without any environment variables

### Workaround
Use the `--pseudo` CLI flag instead:
```bash
lingo.dev run --pseudo
```

But this defeats the purpose of having the config option.

### Proposed Fix
In `setup.ts`, also check the config:
```typescript
const isPseudo = ctx.flags.pseudo || ctx.config?.dev?.usePseudotranslator;
const provider = isPseudo ? "pseudo" : ctx.config?.provider;
```

This ensures both the CLI flag and config option work identically.

ShashwatpSingh:
## Summary
Fixes the issue where `dev.usePseudotranslator: true` in config was ignored during CLI setup, causing unnecessary API key validation.

## Problem
When users set:
```json
{
  "dev": {
    "usePseudotranslator": true
  }
}
```

The CLI still required API keys because it only checked the `--pseudo` CLI flag, not the config option.

## Changes
- Check `ctx.config?.dev?.usePseudotranslator` when selecting the localization provider
- Also check config in authentication skip logic
- Ensures API keys are not required when using pseudotranslator via config

## Testing
- [ ] Verified `usePseudotranslator: true` works without API keys
- [ ] Verified `--pseudo` flag still works
- [ ] Verified normal translation flow still requires API keys

Fixes #1983

## Summary by CodeRabbit

* **New Features**
  * Pseudo-translator can now be enabled through development configuration settings, in addition to the command-line flag option.
  * Authentication step now properly respects the pseudo-translator configuration setting.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -29,7 +29,8 @@ inputs:
     required: false
   api-key:
     description: "Lingo.dev Platform API Key"
-    required: true
+    default: ""
+    required: false
   pull-request:
     description: "Create a pull request with the changes"
     default: false
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,5 +1,17 @@
 # lingo.dev
 
+## 0.131.6
+
+### Patch Changes
+
+- [#1995](https://github.com/lingodotdev/lingo.dev/pull/1995) [`a91ad24`](https://github.com/lingodotdev/lingo.dev/commit/a91ad2427879c561ec4c2a587710b00340104948) Thanks [@Nixxx19](https://github.com/Nixxx19)! - Fix Content-Type header in auth.ts whoami request
+
+## 0.131.5
+
+### Patch Changes
+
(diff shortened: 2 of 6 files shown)


cherkanovart:
Closing in favor of #2019 which includes the same setup.ts changes plus the missing config schema extension (v1.14) needed to make them work at runtime.

PR #2023, opened 2026-03-01
chore(compiler): translation server cli docs update

AleksandrSl:
## Summary

Cleaned up the docs that were left during the development and added an example of how the server url is supposed to be provided

Regarding couple other questions:
- translation-server.log is mentioned in the translations-server readme. But it's nothing more than simple server logs that are always appended to the same file. Initially the file was needed because we started the server in the process that didn't have an access to the console (node process started by turbopack). Now we don't do that, but file still makes sense since it's easier to grep and the amount of logs is big.
- metadata dev vs prod. Metadata is a temporary data describing the pieces of text extracted that need translation. It's used by translation server to get the context and text for translation given the hash of entity to translate which is embedded by compiler into the source code. It doesn't make much sense to keep metadata persisted, it is useful while translation server is working. So we clean it up when we stop the dev process of build. However cleanup is not 100% reliable. So if we make a lot of changes during the dev mode we get a lot of entries in the metadata for the text that no longer exists (we do not have a mechanism to detect and clean up stale entries). In build mode we generate metadata file and translate all of it. So if metadata already had some stale entries they will also be translated but not used. To avoid this dev and prod files were separated to minimize contamination of prod metadata with dev entries in the worst scenario. I think it's mostly obsolete now, given we cleanup metadata on start, but anyway keeping dev and prod files separate makes feel better.
- Pseudo translator. Turned on in the dev config, it's a way 1) to check you font supports something beyond english alphabet 2) avoid translations in dev still seeing that the translations will be made in the places pseudotranslation is applied 3) we can (If I didn't) add empty spaces in between to make the word of different length, so the UI is tested for responsiveness, since the same word vary in length in different languages.

- Language resolvers are in a separate PR

## Changes

- [Key change 1]
- [Key change 2]

## Testing

- [ ] [Describe test 1 - what behavior it validates]
- [ ] [Describe test 2 - what edge case it covers]
- [ ] All tests pass locally

## Visuals

- [ ] Before/after screenshots attached
- [ ] Video demo for interactions (< 30s)

## Checklist

Closes #[issue-number]

## Release Notes

* **New Features**
  * Support for specifying a custom translation server URL in development configuration, allowing users to use an externally-managed server instead of auto-starting one.
  * New environment option for the CLI to explicitly set development or production mode.

* **Documentation**
  * Updated translation server documentation with revised features, API endpoint details, and updated usage examples.

* **Refactor**
  * Simplified CLI by removing automatic bundler detection logic.
  * Improved configuration cleanup and handling to properly sanitize undefined values.

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,5 +1,17 @@
 # lingo.dev
 
+## 0.132.9
+
+### Patch Changes
+
+- [#2027](https://github.com/lingodotdev/lingo.dev/pull/2027) [`bb3ec79`](https://github.com/lingodotdev/lingo.dev/commit/bb3ec797ad0379d99f84f20731f37ab03f759b00) Thanks [@AndreyHirsa](https://github.com/AndreyHirsa)! - Add vNext API support to LingoDotDevEngine. When engineId is provided, the engine routes requests to the vNext API. Refactor vNext localizer to use LingoDotDevEngine from SDK instead of custom inline implementation
+
+- [#2026](https://github.com/lingodotdev/lingo.dev/pull/2026) [`a6aa9f2`](https://github.com/lingodotdev/lingo.dev/commit/a6aa9f2df134de2de706246847dcbeef83b13eec) Thanks [@cherkanovart](https://github.com/cherkanovart)! - Exit with non-zero code when localization tasks fail, so CI/CD pipelines correctly detect partial errors
+
+- Updated dependencies [[`bb3ec79`](https://github.com/lingodotdev/lingo.dev/commit/bb3ec797ad0379d99f84f20731f37ab03f759b00)]:
+  - @lingo.dev/_sdk@0.15.2
+  - @lingo.dev/_compiler@0.11.6
+
 ## 0.132.8
 
 ### Patch Changes
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lingo.dev",
-  "version": "0.132.8",
+  "version": "0.132.9",
   "description": "Lingo.dev CLI",
   "private": false,
   "repository": {
--- a/packages/cli/src/cli/cmd/i18n.ts
+++ b/packages/cli/src/cli/cmd/i18n.ts
@@ -600,6 +600,7 @@ export default new Command()
         await new Promise((resolve) => setTimeout(resolve, 50));
       } else {
         ora.warn("Localization completed with errors.");
+        process.exitCode = 1;
         await trackEvent(email, "cmd.i18n.error", {
           flags,
           ...aggregateErrorAnalytics(
--- a/packages/cli/src/cli/cmd/run/exit-code.spec.ts
+++ b/packages/cli/src/cli/cmd/run/exit-code.spec.ts
@@ -1,0 +1,115 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { CmdRunTask, CmdRunTaskResult } from "./_types";
+import { applyRunExitCode } from "./exit-code";
+
+/**
+ * Tests that the CLI exits with non-zero code when localization tasks fail.
+ * This prevents CI/CD pipelines from silently passing on partial errors.
+ */
+describe("run command - exit code on errors", () => {
+  let originalExitCode: number | undefined;
+
+  beforeEach(() => {
+    originalExitCode = process.exitCode;
+    process.exitCode = undefined;
+  });
+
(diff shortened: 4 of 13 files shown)


packages/new-compiler/src/utils/config-factory.ts
```diff
@@ -90,33 +90,52 @@ function inferPluralizationModel(
 export function createLingoConfig(
   options: PartialLingoConfig & Partial<Pick<LingoConfig, LingoInternalFields>>,
 ): LingoConfig {
+  const cleanOptions = { ...options };
+
+  const cleanObject = (obj: any) => {
```
AleksandrSl, line 95: Allows having explicit undefined values and not overwriting the defaults. Mostly useful for CLI

packages/new-compiler/src/translation-server/translation-server.ts
```diff
@@ -314,7 +314,7 @@ export class TranslationServer {
       this.metadata = await loadMetadata(getMetadataPath(this.config));
       this.logger.debug(
         `Reloaded metadata: ${Object.keys(this.metadata).length} entries`,
-      );
+        );
```
AleksandrSl:
```suggestion
      );
```

packages/new-compiler/src/translation-server/cli.ts
```diff
@@ -139,185 +106,28 @@ function parseModelsString(
   return models;
 }
 
-/**
- * Detect and find bundler config file
- */
-async function findBundlerConfig(
```
AleksandrSl, line 145: These all functions were added while I was adding a kind of hidden `_lingoConfig` property to the config, but all the bundlers were unhappy about this showing warning. So I removed it but forgot to clean up this code.
Currently CLI arguments are the only way to configure the server if you run it manually.

packages/new-compiler/src/translation-server/cli.ts
```diff
@@ -1,52 +1,7 @@
 #!/usr/bin/env node
-/**
```
AleksandrSl, line 2: Duplicates help

packages/new-compiler/src/plugin/next.ts
```diff
@@ -210,31 +210,39 @@ export async function withLingo(
...
-    if (translationServer.server) {
-      // We start the server in the same process, so we should be fine without any sync cleanup. Server should be killed with the process.
-      registerCleanupOnCurrentProcess({
-        asyncCleanup: async () => {
-          await translationServer.server.stop();
+    if (lingoConfig.dev.translationServerUrl) {
+      logger.info(`Using existing translation server URL (${lingoConfig.dev.translationServerUrl}) from config`);
+      process.env.LINGO_TRANSLATION_SERVER_URL =
+        lingoConfig.dev.translationServerUrl;
+    } else {
```
AleksandrSl, line 217: This is the only real change here, it was lost during development. If we have the url provided, we don't need to start anything

packages/new-compiler/src/translation-server/cli.ts
```diff
@@ -342,34 +153,23 @@ Options:
...
   --use-pseudo                 Use pseudotranslator (default: false)
+  --env <string>               Environment: "development" or "production" (default: "development")
   --config <path>              Path to config file (JSON)
   --help                       Show this help message
 
 Examples:
+  # Run script from the project root where next or vite configuration is present.
   # Start with defaults
-  npx @lingo.dev/compiler/translation-server
+  pnpm tsx ../../packages/new-compiler/src/translation-server/cli.ts --port 3456 --target-locales "es,fr,de,ja"
```
AleksandrSl, line 163: No really clean way to start the server, but since it's only for internal debugging it should be fine

packages/new-compiler/src/translation-server/translation-server.ts
```diff
@@ -314,7 +314,6 @@ export class TranslationServer {
       this.metadata = await loadMetadata(getMetadataPath(this.config));
       this.logger.debug(
         `Reloaded metadata: ${Object.keys(this.metadata).length} entries`,
-      );
```
AndreyHirsa, line 317: It looks like the closing parenthesis was accidentally removed here. Could you please add it back?

AleksandrSl: oops

AndreyHirsa:
@AleksandrSl Looks good, thank you. Can you please add a changeset?

AleksandrSl:
@AndreyHirsa do we need it for the docs update? If there is no way to skip it, I'll add one 😄

AndreyHirsa:
@AleksandrSl Yeah, we need a changeset for every change.

PR #2024, opened 2026-03-01
feat(compiler): custom resolvers

AleksandrSl:
## Summary

It's an example of having custom resolvers and what we have to do to support them. Ofc we have to make path based i18n support internally along with the cookies, without custom resolvers.

There are both changes on the user side (demo) and our side (in the compiler, because I removed the custom resolver part for simplicity until we have requests, but PR also demonstrates how we can support path based i18n, probably it can also be done a bit cleaner, and we can provide middleware configuration as exported function, so users don't have to do it, similar how next-intl does this. suprisingly they have a ton of code around these)

## Changes

- [Key change 1]
- [Key change 2]

## Testing

- [ ] [Describe test 1 - what behavior it validates]
- [ ] [Describe test 2 - what edge case it covers]
- [ ] All tests pass locally

## Visuals

- [ ] Before/after screenshots attached
- [ ] Video demo for interactions (< 30s)

## Checklist

Closes #[issue-number]

## Summary by CodeRabbit

* **New Features**
  * Added path-based locale routing through URL segments (e.g., /en/, /es/, /de/).
  * Implemented automatic locale detection from browser language preferences.
  * Enabled custom locale resolvers for flexible locale-switching behavior.
  * Enhanced navigation with locale-aware link handling.


packages/new-compiler/src/index.ts
```diff
@@ -1 +1,2 @@
 export type { PartialLingoConfig } from "./types";
+export type { LocaleCode } from "lingo.dev/spec";
```
AleksandrSl, line 2: Makes sense to reexport I think, because it's used in many places, and compiler could be the single dependency.

packages/new-compiler/src/virtual/code-generator.ts
```diff
@@ -57,5 +71,9 @@ export function persistLocale(locale) {
     document.cookie = \`${cookieName}=\${locale}; path=/; max-age=${maxAge}\`;
   }
 }
+
+export function getLocalePathname(locale) {
+  return null; // Not used for cookie-based routing
+}
```
AleksandrSl, line 77:
```suggestion
```

packages/new-compiler/src/virtual/code-generator.ts
```diff
@@ -39,6 +46,13 @@ export const getServerLocale = createNextCookieLocaleResolver({ cookieConfig: ${
  * Exports getClientLocale() and persistLocale() functions
  */
 export function generateClientLocaleModule(config: LingoConfig): string {
+  if (config.localePersistence.type === "custom") {
+    // For custom resolvers, import from an abstract path that will be
+    // resolved by Turbopack's resolveAlias to the actual user file
+    return `export { getClientLocale, persistLocale, getLocalePathname } from '@lingo.dev/compiler/virtual/locale-client';`;
```
AleksandrSl:
```suggestion
    return `export { getClientLocale, persistLocale } from '@lingo.dev/compiler/virtual/locale-client';`;
```

packages/new-compiler/src/react/shared/LingoProvider.tsx
```diff
@@ -178,22 +179,26 @@ function LingoProvider__Prod({
...
-      // 2. Update local state for immediate UI feedback
+      // Update local state for immediate UI feedback
       setLocaleState(newLocale);
 
-      // 3a. Next.js pattern: Trigger server re-render
+      // Next.js pattern: Trigger server re-render
       if (router) {
-        router.refresh();
+        if (newUrl) {
+          router.push(newUrl);
```
AleksandrSl, line 196: I missed that for the path based i18n we will have to do navigation. I don't think we need to add a separate method for that, just returning newUrl form `persistLocale` if redirect is needed sounds reasonable to me.

Next needs a `push`, because we have to trigger a middleware so that server components get a correct locale

packages/new-compiler/src/plugin/resolve-locale-resolver.ts
```diff
@@ -0,0 +1,89 @@
...
+  sourceRoot: string,
+  lingoDir: string,
+  projectRoot: string,
+): {
+  serverResolver: string;
+  clientResolver: string;
+} {
+  const baseDir = path.join(projectRoot, sourceRoot, lingoDir);
+
+  const serverPath = resolveResolverPath("locale-resolver-server", baseDir);
```
AleksandrSl, line 69: The whole file is a way to check that the file are present where we expect them to be. In this PR I expect them to be in the lingo directory, but we are free to come up with any pattern we want.

Also a nice error if one of the files is missing when custom locale resolver is chosen in settings would be good

packages/new-compiler/src/plugin/next.ts
```diff
@@ -259,9 +260,31 @@ export async function withLingo(
...
+  //
+  // Convention: Resolver files must be at <sourceRoot>/<lingoDir>/locale-resolver-{server|client}.ts
+  if (lingoConfig.localePersistence.type === "custom") {
+    const resolvedPaths = resolveCustomResolverPaths(
+      lingoConfig.sourceRoot,
+      lingoConfig.lingoDir,
+      process.cwd(),
+    );
+
+    customResolverAliases = {
```
AleksandrSl, line 279: For vite and unplugin there should already be a similar thing.

See 
```
function tryLocalOrReturnVirtual(
  config: LingoConfig,
  fileName: string,
  virtualName: string,
) {
  const customPath = path.join(config.sourceRoot, config.lingoDir, fileName);
  if (fs.existsSync(customPath)) {
    return customPath;
  }
  return virtualName;
}
```

I don't remember if I checked it working

demo/new-compiler-next16/supported-locales.ts
```diff
@@ -0,0 +1,4 @@
+import type { LocaleCode } from "@lingo.dev/compiler"
```
AleksandrSl, line 1: Just  a way to avoid repeating these

demo/new-compiler-next16/proxy.ts
```diff
@@ -0,0 +1,107 @@
+import { NextRequest, NextResponse } from "next/server";
```
AleksandrSl, line 1: It's 2026 why the hell Next asks users to configure all this hell when they want path based i18n

PR #2048, opened 2026-03-18
fix(new-compiler): disable noSync for shared metadata LMDB writes

### LINKED ISSUE #2047: @lingo.dev/compiler 0.4.0 intermittently crashes Turbopack builds (evaluate_webpack_loader / failed to receive message)
## Update

Root cause has now been confirmed in the upstream repository and a fix PR is open:
- PR: #2048

The confirmed trigger is concurrent metadata writes to the same LMDB path with `noSync=true` in the compiler metadata pipeline. In local stress testing, that combination caused worker aborts (`exit 134`) and, in some runs, missing metadata entries without a crash.

## Summary

Upgrading `@lingo.dev/compiler` from `0.1.4` to `0.4.0` introduces intermittent Turbopack production build panics in a Next.js app that uses the Next integration (`withLingo(...)`).

The failure is random but reproducible with repeated builds. The panic consistently points at the webpack-loader execution path inside Turbopack:

```text
Error [TurbopackInternalError]: failed to receive message

Caused by:
- reading packet length
- unexpected end of file

Debug info:
- Execution of get_all_written_entrypoints_with_issues_operation failed
- Execution of PlainIssue::from_issue failed
- Execution of PlainSource::from_source failed
- Execution of <WebpackLoadersProcessedAsset as Asset>::content failed
- Execution of WebpackLoadersProcessedAsset::process failed
- Execution of evaluate_webpack_loader failed
```

## Environment

- `@lingo.dev/compiler`: `0.4.0`
- Next.js: `16.1.7`
- React: `19.2.4`
- Node.js: `24.14.0`
- pnpm: `10.32.1`
- macOS (Apple Silicon)

The app uses the standard Next integration:

```ts
import { withLingo } from "@lingo.dev/compiler/next";

export default async function createNextConfig() {
  return await withLingo(nextConfig, {
    sourceRoot: "src",
    lingoDir: ".lingo",
    sourceLocale: "en",
    targetLocales: [...],
    buildMode: "cache-only",
    pluralization: { enabled: false, model: ... },
  });
}
```

Build command:

```bash
LINGO_BUILD_MODE=cache-only next build --turbo
```

## What I verified

I isolated this in a real app by checking historical commits and then testing package combinations in a detached worktree.

### Commit-level regression window

- pre-upgrade commit (`fc5c5af`): `0/8` Turbopack panics
- upgrade commit (`0701760`): `4/8` Turbopack panics

### Package isolation

On top of the regressing commit:

- baseline (`@lingo.dev/compiler 0.4.0`, `next 16.1.7`): `3/6` panics
- only downgrade `@lingo.dev/compiler` to `0.1.4` while keeping newer Next/tooling: `0/6` panics
- only downgrade `next` to `16.1.1` while keeping `@lingo.dev/compiler 0.4.0`: `3/6` panics

So from my testing, `Next` is not the primary trigger. The regression follows the Lingo compiler upgrade.

### Source-code isolation

I also reverted the business/source changes from the regressing commit while keeping the upgraded dependency set. The panic still reproduced, so this does not look like a specific app component bug.

## Strong suspicion

I compared the package internals between `0.1.4` and `0.4.0`.

The biggest implementation difference relevant to this failure is that `0.4.0` appears to have changed metadata handling substantially:

- `0.1.4` uses JSON metadata files (`metadata-build.json` / `metadata-dev.json`)
- `0.4.0` uses LMDB-backed metadata storage and a different metadata write path during loader execution

Because the Turbopack panic consistently happens in `evaluate_webpack_loader`, my current hypothesis is:

> the `0.4.0` Next/Turbopack loader path is sometimes crashing its worker/process during metadata writes, and Turbopack only surfaces that as `failed to receive message` / EOF.

I cannot prove the exact crash line without instrumenting the package, but the package-level regression and the metadata implementation change line up very closely.

## Additional observation

I have another smaller Next app using `@lingo.dev/compiler 0.4.0` that does **not** reproduce this issue in repeated builds. My current interpretation is that this is workload-sensitive: the larger app has a bigger route graph and many more translatable entries, so it triggers the issue much more reliably.

For reference:

- app that reproduces: ~248 TSX files, ~216 app files, ~1625 translatable entries
- smaller app that does not reproduce: ~179 TSX files, ~118 app files, ~991 translatable entries

## Expected behavior

Repeated `next build --turbo` runs should be stable when using `withLingo(...)`.

## Actual behavior

Production Turbopack builds intermittently panic with `failed to receive message` / `evaluate_webpack_loader failed`.

## Request

Please investigate whether the `0.4.x` Next/Turbopack integration is unsafe under concurrent loader execution, especially around metadata persistence and LMDB usage.

If helpful, I can also prepare a minimized reproduction derived from this app, but the matrix above already narrowed the regression to the Lingo compiler upgrade itself.

visoar:
## Linked issue
- Root-cause investigation and original report: #2047

## Summary
- force synchronized LMDB commits for compiler metadata writes
- add a multi-process regression test for `saveMetadata(..., true)`

## Root cause
`@lingo.dev/compiler` writes build metadata from multiple bundler worker processes into the same LMDB path.

Before this patch, `saveMetadata()` forwarded `noSync=true` during production builds. In local stress testing on macOS, that combination caused two failure modes:

1. worker processes aborting with exit code `134`
2. successful exits with missing metadata entries

The failures reproduced without Next.js or app code involved, using only concurrent child processes calling `saveMetadata()` against the same database path.

The same stress test passed when:
- each worker wrote to its own LMDB path, or
- the shared metadata database used `noSync=false`

That points to `noSync=true` on a shared metadata LMDB as the unsafe part of the current implementation.

## Validation
- `pnpm --filter @lingo.dev/compiler exec vitest --run src/metadata/manager.test.ts`
- repeated 24-process stress runs writing 2400 total entries with `saveMetadata(..., true)` now finish at `2400/2400`

## Notes
I also saw unrelated existing workspace issues when running the full package test/typecheck suite from a fresh clone:
- missing `lingo.dev/spec` / `lingo.dev/sdk` resolution in `tsc`
- existing vitest failures in `src/react/server-only/index.test.tsx` and `src/translation-server/translation-server.test.ts`

Those were present outside this patch, so this PR keeps the scope limited to the metadata regression.

Closes #2047

## Summary by CodeRabbit

* **New Features**
  * Added multi-process concurrent metadata handling with synchronized data writes.

* **Tests**
  * Introduced concurrent access testing to verify metadata integrity across multiple parallel processes.

### DIFF
--- a/packages/new-compiler/src/metadata/fixtures/save-metadata-worker.ts
+++ b/packages/new-compiler/src/metadata/fixtures/save-metadata-worker.ts
@@ -4,7 +4,29 @@ import { generateTranslationHash } from "../../utils/hash";
 const dbPath = process.argv[2];
 const workerId = Number(process.argv[3]);
 const iterations = Number(process.argv[4]);
-const noSync = process.argv[5] === "true";
+const noSyncArg = process.argv[5];
+
+if (!dbPath) {
+  console.error("Missing dbPath argument");
+  process.exit(2);
+}
+
+if (Number.isNaN(workerId)) {
+  console.error("Invalid workerId argument");
+  process.exit(2);
+}
+
+if (Number.isNaN(iterations)) {
+  console.error("Invalid iterations argument");
+  process.exit(2);
+}
+
+if (noSyncArg !== "true" && noSyncArg !== "false") {
+  console.error("Invalid noSync argument");
+  process.exit(2);
+}
+
+const noSync = noSyncArg === "true";
 
 function createEntry(i: number) {
   const sourceText = `worker-${workerId}-entry-${i}`;
--- a/packages/new-compiler/src/metadata/manager.test.ts
+++ b/packages/new-compiler/src/metadata/manager.test.ts
@@ -70,6 +70,13 @@ async function runSaveMetadataWorker(
     child.stderr.on("data", (chunk) => {
       stderr += chunk.toString();
     });
+    child.on("error", (error) => {
+      reject(
+        new Error(
+          `worker ${workerId} failed to spawn: ${error.message}\nstdout:\n${stdout}\nstderr:\n${stderr}`,
+        ),
+      );
+    });
 
     child.on("exit", (code, signal) => {
       if (code === 0) {
--- a/packages/new-compiler/src/metadata/manager.ts
+++ b/packages/new-compiler/src/metadata/manager.ts
@@ -7,6 +7,8 @@ import { logger } from "../utils/logger";
 
 const METADATA_DIR_DEV = "metadata-dev";
 const METADATA_DIR_BUILD = "metadata-build";
+// Metadata writes require synchronized commits because bundlers can fan out
+// multiple worker processes against the same LMDB path.
 const METADATA_DB_NO_SYNC = false;
 
 /**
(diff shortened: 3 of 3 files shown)


visoar:
Addressed the review feedback in aa829ee8:\n- simplified the synchronized metadata flag handling\n- added CLI argument validation in the worker fixture\n- added a spawn error handler in the multi-process regression test\n\nRe-ran: `pnpm --filter @lingo.dev/compiler exec vitest --run src/metadata/manager.test.ts`

PR #2055, opened 2026-03-24
feat(cli): add --batch-size parameter to prevent context leaking

### LINKED ISSUE #1733: Add batch size parameter to prevent context leaking of translator notes/hints
### Problem

This is a follow up to https://github.com/lingodotdev/lingo.dev/issues/1728, since hints currently are not included in the request to a model at all (not when using the lingo.dev platform I think)

Given the following example:
```jsonc
{
  // Short form for "Year", used to display a date format, example in `en`: "DD-MM-YYYY". Don’t translate for CJK (Chinese, Japan, Korean) languages
  "DATE_FORMAT_SPECIFIER_YEAR": "Y",
  // Short form for "Month", used to display a date format, example in `en`: "DD-MM-YYYY"
  "DATE_FORMAT_SPECIFIER_MONTH": "M"
}
```

in my tests, using `openai/gpt-oss-120b` (openAI compatible provider, see https://github.com/lingodotdev/lingo.dev/issues/1729) there is some context leaking going on, even when trying to use a prompt like:

> Translate the provided text from source: {source} to target: {target}. For each translation key, use only the hint with identical key. If no hint exists for a key, *do not* reuse any other hints

and translate to korean, the result it get is:
```
{
  "DATE_FORMAT_SPECIFIER_YEAR": "Y",
  "DATE_FORMAT_SPECIFIER_MONTH": "M"
}
```
even though only `DATE_FORMAT_SPECIFIER_YEAR` has the hint to not translate to CJK languages this get’s applied to `DATE_FORMAT_SPECIFIER_MONTH` aswell.

Looking at the reasoning it becomes clear, that the model is not able to follow the instruction to treat each hint for each translation completely individually. I tried many things, but could not find a reliable solution.

The way out is probably to translate each string one-by-one in individual requests to the LLM, preventing context leaking between translations completely. This of course has other trade-offs (increased API cost, slower, …) but it reliably translates to:

```
{
  "DATE_FORMAT_SPECIFIER_YEAR": "Y",
  "DATE_FORMAT_SPECIFIER_MONTH": "월"
}
```

### Solution

Add a batch size parameter to the CLI to control how many translations strings are send to a LLM at once. A batch size of `1` would send each translation individually.

### Visuals

Image

### Workarounds

_No response_

Hellnight2005:
## Summary

Implements the `--batch-size` parameter for CLI commands to control the number of translation strings sent to the LLM per request, preventing context leaking of translator hints between unrelated keys.

## Changes

- Added `--batch-size <number>` flag to the `run` and `i18n` CLI commands.
- Updated `createBasicTranslator` in `src/cli/processor/basic.ts` to break down and process inputs into configurable chunks.
- Updated [createExplicitLocalizer](cci:1://file:///c:/Users/Abhijeet/Desktop/lingo.dev/packages/cli/src/cli/localizer/explicit.ts:15:0-102:1) in [src/cli/localizer/explicit.ts](cci:7://file:///c:/Users/Abhijeet/Desktop/lingo.dev/packages/cli/src/cli/localizer/explicit.ts:0:0-0:0) to execute batches properly and consolidate the chunked responses.
- Fixed a bug causing the pseudo-localizer to crash by implementing the missing [validateSettings](cci:1://file:///c:/Users/Abhijeet/Desktop/lingo.dev/packages/cli/src/cli/localizer/explicit.ts:174:4-193:5) method.
- Handled edge cases for `JSON.parse` exceptions gracefully using `jsonrepair` in the AI feedback loop. 

## Testing

- [x] Verified chunking logic correctly isolates inputs when `batchSize` is strictly set to 1.
- [x] Verified the default behavior reliably places smaller files into a single batch without chunking unnecessary data boundaries.
- [x] All tests pass locally (verified the 846 unit tests in `packages/cli` succeed).

## Visuals

- [N/A] Before/after screenshots attached *(CLI backend behavior only)*
- [N/A] Video demo for interactions (< 30s)

## Checklist

- [x] Tests cover business logic (not just happy path)
- [x] No breaking changes (or documented below)

Closes #1733

## Summary by CodeRabbit

* **New Features**
  * Added `--batch-size` CLI option to `run` and `i18n` commands, allowing users to customize translation batch processing with flexible batch sizes between 1 and 250
  * Implemented intelligent chunked processing during AI-driven localization workflows with real-time progress reporting for each batch being processed

* **Tests**
  * Added comprehensive test coverage for batch processing behavior and correctness


packages/cli/src/cli/processor/basic.ts
```diff
@@ -106,9 +111,11 @@ function extractPayloadChunks(
     currentChunkItemCount++;
 
     const currentChunkSize = countWordsInRecord(currentChunk);
+    const effectiveBatchSize =
```
AndreyHirsa: Now, when `--batch-size` isn’t provided, the effective batch size defaults to the total number of keys. I would keep 25 as the fallback instead of payloadEntries.length.

packages/react/src/client/index.ts
```diff
@@ -5,3 +5,4 @@ export * from "./component";
 export * from "./locale-switcher";
 export * from "./attribute-component";
 export * from "./locale";
+export { getLocaleFromCookies, setLocaleInCookies } from "./utils";
```
AndreyHirsa: This doesn't seem related to batch size

packages/cli/src/cli/cmd/run/_types.ts
```diff
@@ -56,5 +56,6 @@ export const flagsSchema = z.object({
   debounce: z.number().positive().prefault(5000), // 5 seconds default
   sound: z.boolean().optional(),
   pseudo: z.boolean().optional(),
+  batchSize: z.number().min(1).optional(),
```
AndreyHirsa: Could you please add a maximum value of 250?

packages/cli/src/cli/cmd/i18n.ts
```diff
@@ -662,6 +666,7 @@ function parseFlags(options: any) {
     file: Z.array(Z.string()).optional(),
     interactive: Z.boolean().prefault(false),
     debug: Z.boolean().prefault(false),
+    batchSize: Z.number().min(1).optional(),
```
AndreyHirsa: Same as run: use a maximum value of 250.

packages/cli/src/cli/cmd/run/setup.ts
```diff
@@ -54,7 +54,12 @@ export default async function setup(input: CmdRunContext) {
             ctx.flags.pseudo || ctx.config?.dev?.usePseudotranslator;
           const provider = isPseudo ? "pseudo" : ctx.config?.provider;
           const engineId = ctx.config?.engineId;
-          ctx.localizer = createLocalizer(provider, engineId, ctx.flags.apiKey);
+          ctx.localizer = createLocalizer(
```
AndreyHirsa, line 57: Please use prettier to match the project formatting.

packages/cli/src/cli/localizer/explicit.ts
```diff
@@ -183,85 +192,205 @@ function createAiSdkLocalizer(params: {
...
   };
 }
+
+/**
+ * Extract payload chunks based on the ideal chunk size
+ * @param payload - The payload to be chunked
+ * @param batchSize - Max number of keys per chunk (default: 25)
+ * @returns An array of payload chunks
+ */
+function extractPayloadChunks(
```
AndreyHirsa: `extractPayloadChunks` and `countWordsInRecord` are now copy-pasted in two places  - `packages/cli/src/cli/processor/basic.ts` and at the bottom of `packages/cli/src/cli/localizer/explicit.ts`. Please extract them into a shared utility so we don't maintain two copies.

packages/cli/src/cli/localizer/explicit.ts
```diff
@@ -183,85 +192,205 @@ function createAiSdkLocalizer(params: {
...
+        }
+        let finalResult: Record<string, any> = {};
 
-      // Handle both object and string responses
-      if (typeof result.data === "object" && result.data !== null) {
-        return result.data;
+        // Handle both object and string responses
+        if (typeof result?.data === "object" && result.data !== null) {
+          finalResult = result.data;
+        } else if (result?.data) {
```
AndreyHirsa: What are the problems with the current repair logic? This looks overly complex for what it does.

AndreyHirsa (changes requested):

packages/cli/src/cli/cmd/run/setup.ts
Hellnight2005: I looked into this and tried reformatting it. I also cross-verified it (including with AI and Prettier), and it seems the line is being split due to the `printWidth: 80` rule in the project config.

packages/cli/src/cli/localizer/explicit.ts
Hellnight2005: The original repair logic was added to handle inconsistencies across different LLM providers. In some cases, models return the `data` field as double-stringified JSON (e.g. `{"data": "{\"message\": \"Hola\"}"}`) instead of an object.

The manual extraction (`indexOf('{') `+ `jsonrepair`) was intended to safely recover that payload and prevent crashes during batch localization.

That said, I agree the implementation was overly complex. I’ve simplified it by removing the manual slicing and reusing `parseModelResponse()` recursively, which handles the double-stringification more cleanly with less code.

packages/cli/src/cli/localizer/explicit.ts
Hellnight2005: I’ve removed the duplicate implementations from `processor/basic.ts` and `localizer/explicit.ts`

I extracted them into a shared utility (`packages/cli/src/cli/utils/chunk.ts`) and imported them back into both places so we only maintain them in one place

PR #2142, opened 2026-06-22
fix(deps): drop pkg

moygospadin:
## Summary by CodeRabbit

* **Chores**
  * Removed `external-editor` from production dependencies and moved `gray-matter` to development dependencies to address remaining audit items.
* **Refactor**
  * Improved the deprecated i18n CLI command’s interactive editor workflow with safer temp-file handling and more robust editor invocation.
  * Replaced `gray-matter`-based `---` frontmatter parsing/serialization in markdown and MDX loaders with shared custom helpers.


maxprilutskiy (approved):

moygospadin:
Addressed the two CodeRabbit findings:

**1. `open-editor.ts` editor parsing — fixed.** Switched from `editor.split(/\s+/)` to `spawnSync(\`${editor} "${file}"\`, { shell: true })`, so multi-word/quoted editor commands (e.g. `EDITOR='"/path with spaces/code" --wait'`) tokenize correctly. Both inputs are trusted (`editor` = user's own env var, `file` = our `mkdtemp` path), so no injection surface. No new dependency added.

**2. `_frontmatter.ts` regex boundary — skipped (verified false positive).** The suggestion (require `(?:\r?\n|$)` after the closing `---`) would actually *diverge* from `gray-matter`, which this helper must match (the loader specs assert equivalence). Verified against gray-matter:

| input | gray-matter | current `\r?\n?` | proposed `(?:\r?\n|$)` |
|---|---|---|---|
| `---\na: 1\n---foo` | `data={a:1}`, `content="foo"` | ✅ same | ❌ no match (diverges) |
| `---\na: 1\n---` (EOF) | `content=""` | ✅ same | ✅ same |
| `---\na: 1\n---\n\n# H` | `content="\n# H"` | ✅ same | ✅ same |

gray-matter does **not** require a newline after the closing `---`, so the existing `\r?\n?` is the behavior-matching choice; the lookahead would break the gray-matter equivalence the specs guarantee. Loader specs (60) remain green.

PR #2112, opened 2026-06-13
fix(spec): improve config semantic validation errors

jaya6400:
Added validateConfigSemantics() that runs on successful parse and before the upgrade fallback, with helpers for legacy-upgrade detection and unknown bucket ID resolution (v1 path→type vs v1.1+ type→config formats).

New error messages:

- Missing source: Source locale is required. Add "locale.source" to your i18n.json (e.g. "en").
- Empty targets: At least one target locale is required. Add locale targets to your i18n.json (e.g. ["es", "fr"]).
- Duplicate targets: Duplicate target locale(s): es. Remove duplicates from locale.targets.
- No buckets: No buckets configured. Add at least one bucket to i18n.json under "buckets"
- Unknown bucket types: Unknown bucket type(s): invalid, not-a-bucket. Supported types: ail, android, csv.
Also added .min(1) on localeSchema.targets as schema-level defense.

Tests: All 35 spec tests pass (pnpm --filter "@lingo.dev/_spec" test).

## Testing

- [x] All tests pass locally
Capture

## Visuals

**Required for UI/UX changes:**

- [ ] Before/after screenshots attached
- [ ] Video demo for interactions (< 30s)

## Checklist

- [x] Tests cover business logic (not just happy path)
- [x] No breaking changes (or documented below)

Closes [#561](https://github.com/lingodotdev/lingo.dev/issues/561)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved i18n configuration validation error messages with clearer, actionable guidance for common issues: missing `locale.source`, missing/empty `locale.targets`, duplicate target locales, missing/empty `buckets`, and unknown bucket types.
  * Validation is now stricter for `locale.targets`, requiring at least one target locale and surfacing a dedicated “at least one target locale is required” message when omitted.
  * Semantic validation now aggregates multiple configuration problems into a single, readable error.

PR #2144, opened 2026-06-23
feat(spec): improve error messages for invalid locale codes

### LINKED ISSUE #561: Spec Module (/packages/spec): Improve error messages in configuration validation
In `config.ts`, when validation fails, error messages could be more descriptive to help users troubleshoot configuration issues. The error handling in the `parse` method could provide more context about what specifically is wrong with the configuration.

jaya6400:
### Summary
- This PR improves the error message shown when a user specifies an unsupported or invalid locale code in their i18n.json config. Instead of a generic Zod validation error, the parser now extracts and displays the exact invalid locale value, making the error more actionable.

- Crucially, this PR does not change any validation behavior – only the error message is improved. The parser continues to throw only for invalid locale codes (same as main), and all other validation issues (missing source, unknown bucket types, etc.) still attempt an upgrade.

### Changes:
- Added isLocaleError helper to cleanly identify Zod issues related to invalid locale codes.

- Added getValueAtPath helper to safely extract the actual invalid locale value from the raw config object.

- Updated the parse function in extendConfigDefinition to build error messages that include the unsupported locale(s), e.g.:

```
Unsupported locale: xxxx
```
instead of a generic Zod message.

- Removed the mistakenly added missing‑source hard error (which would have been a breaking change) – validation logic now matches main exactly.

### Testing
Logic tests added:

- should throw an error for unsupported locales – verifies that multiple invalid locales are reported in the error message with their values.

- should throw for invalid locale codes with clear error – confirms that a single invalid locale displays the exact value.

- should parse the default config without errors – ensures the built‑in default config is still accepted.

- Existing upgrade tests (v0 → latest, v1 → latest) pass, verifying backward compatibility.

All 34 unit tests in @lingo.dev/_spec pass locally.
image

### Checklist:
- Changeset added – included as @lingo.dev/_spec patch bump.

- Tests cover business logic (not just happy path) – the added tests explicitly cover invalid locale edge cases and ensure no regression.

- No breaking changes – validation behavior is identical to main. Only the error message is improved.

Closes #561 

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid locale code error messages now display the unsupported locale value, providing clearer feedback when configuration validation fails.


jaya6400:
Hey @moygospadin, I've reworked the changes based on your feedback from https://github.com/lingodotdev/lingo.dev/pull/2112#issuecomment-4780711914. This new PR (#2144) only improves error messages and removes all the strict validation rules that were causing breaking changes. The default config now parses correctly, and the error string duplication is fixed. 
Please review the changes, let me know if any further changes required. Thanks!
@cherkanovart @maxprilutskiy

jaya6400:
Hi @moygospadin , I see one CI check is failing, I checked it is a pre existed failure on "main" branch as well not related to my changes. 
Is there any update on this PR review, do I need to make any further changes?
@maxprilutskiy @cherkanovart @AndreyHirsa

AndreyHirsa:
Hey @jaya6400, thanks for sticking with this! I dug into the current diff and it actually comes out behavior-identical to `main` — it already throws `Unsupported locale: <value>` on this path, so this version ends up being a refactor rather than a new message (and a few of the tests overlap with the existing `upgrade v0/v1` / `unsupported locales` ones). I'll leave the PR open, but rather than iterate on code, let's first agree on *what* should improve over on #561 — which cases to cover and the exact messages, kept non-breaking. Once we're aligned I'll happily review the change here. 🙏

PR #2171, opened 2026-07-19
fix(compiler): use a target-language few-shot example to prevent cros…

### LINKED ISSUE #2041: compiler: i18n translations can be generated in Spanish when using a custom OpenAI-compatible model
## Description

We hit a translation correctness issue in `@lingo.dev/compiler` while building a Next.js app with `sourceLocale: "en"` and `targetLocales: ["zh-Hans"]`.

In our setup, newly generated `zh-Hans` translations were sometimes written in Spanish instead of Simplified Chinese.

For example, these strings were generated into `src/.lingo/cache/zh-Hans.json` and `.next/zh-Hans.json` as:

- `Uploaded` -> `Subido`
- `Drag, drop, or browse` -> `Arrastre, suelte o explore`
- `Queued` -> `En cola`
- `Needs attention` -> `Necesita atención`
- `Add images` -> `Agregar imágenes`
- `Canceled` -> `Cancelado`

The target locale in the request was definitely `zh-Hans`, not `es`.

## Environment

- `@lingo.dev/compiler`: `0.4.0`
- Next.js: `16.1.6`
- React: `19`
- App type: Next.js App Router
- Build mode: `translate`
- Target locale: `zh-Hans`

## Configuration

`lingo.config.ts`:

```ts
import { SOURCE_LOCALE, TARGET_LOCALES } from "./src/lib/config/i18n";

export const LINGO_DEFAULT_MODEL = "openai:LongCat-Flash-Chat";

export const LINGO_MODEL_MAP: Record<string, string> = {
  "*:*": LINGO_DEFAULT_MODEL,
};

export const LINGO_PLURALIZATION_MODEL = LINGO_DEFAULT_MODEL;
```

Environment:

```bash
OPENAI_BASE_URL=https://api.longcat.chat/openai/v1
OPENAI_API_KEY=***
```

So this is using the `openai:` provider with an OpenAI-compatible custom base URL.

## Steps to Reproduce

1. Configure Lingo compiler with:
   - `sourceLocale: "en"`
   - `targetLocales: ["zh-Hans"]`
2. Use a custom OpenAI-compatible model via:
   - `models: { "*:*": "openai:LongCat-Flash-Chat" }`
   - `OPENAI_BASE_URL=https://api.longcat.chat/openai/v1`
3. Add a few new English UI strings that are not yet present in `.lingo/cache/zh-Hans.json`
4. Run production build with translation generation enabled
5. Inspect newly generated entries in:
   - `.lingo/cache/zh-Hans.json`
   - `.next/zh-Hans.json`

## Expected Result

New translations for `zh-Hans` should always be generated in Simplified Chinese.

## Actual Result

Some newly generated `zh-Hans` entries were generated in Spanish.

## What We Verified

We checked the following locally:

- The app only supports `en` and `zh-Hans`
- `targetLocales` is correct
- Lingo translation requests were sent to `/translations/zh-Hans`
- The wrong-language output only affected newly generated cache misses
- Existing cached Chinese entries remained correct

This suggests the locale routing is correct, but the final model output can drift into Spanish under this setup.

## Possible Cause

This may be caused by a combination of:

1. using a custom OpenAI-compatible model/backend
2. the compiler's built-in few-shot examples being `en -> es`

In `@lingo.dev/compiler@0.4.0`, the bundled few-shot example in `build/translators/lingo/shots.mjs` is Spanish-based (`locale: "es"`). A weaker or less instruction-following OpenAI-compatible model may get biased by that example and respond in Spanish even when the target locale is `zh-Hans`.

## Suggested Improvements

Any of these would likely help:

1. Make the few-shot examples target-locale aware instead of always using Spanish
2. Add a validation step that rejects obviously wrong-language output for the requested locale
3. Allow disabling bundled few-shot examples for custom model setups
4. Document that some weaker OpenAI-compatible models may produce cross-locale drift

## Additional Notes

We manually corrected the wrong `zh-Hans` cache entries and confirmed the app then rendered correctly.

If helpful, I can provide a minimal reproduction config based on Next.js 16 + `@lingo.dev/compiler@0.4.0`.

adityasharmawork:
Fixes cross-locale drift in the compiler's LLM translator: with a custom OpenAI-compatible model, newly generated translations could come out in Spanish regardless of the requested target locale, because the bundled few-shot example was always English→Spanish.

## Changes

* `translators/lingo/shots.ts`: replaced the single hard-coded English→Spanish few-shot with a `getShots(targetLocale)` selector that returns only an example whose language matches the target. Kept the existing Spanish example and added correct ones for a curated set (fr, de, it, pt, nl, ru, ja, ko, and Simplified/Traditional Chinese); any other target gets no example rather than a wrong-language one. Chinese matching is script-aware (Simplified vs Traditional).
* `translators/lingo/translator.ts`: inject `getShots(targetLocale)` instead of the static `shots` array.
* `translators/lingo/prompt.ts`: name the source/target language in the built-in system prompt via `Intl.DisplayNames` (e.g. `Target language (locale code): zh-Hans (Simplified Chinese)`), with a graceful fallback to the bare code.
* Added a changeset (patch bump for `@lingo.dev/compiler`).

## Testing

- [X] `shots.test.ts` — `getShots()` returns the example matching the target language and never a mismatched one: Spanish variants (`es`, `es-419`, `es-MX`) → Spanish; script-aware Chinese (`zh-Hans`/`zh-CN`/bare `zh` → Simplified, `zh-Hant`/`zh-TW` → Traditional); `ja`/`ko`/`fr-CA` → their language.
- [X] `shots.test.ts` — uncovered locales and the source language (`sw`, `hu`, `en`) return no example (regression guard against wrong-language drift), and each example keeps English input with output keys aligned to input.
- [X] `prompt.test.ts` — the built-in prompt names both languages (`en (English)`, `es (Spanish)`, `zh-Hans (Simplified Chinese)`), and the user-defined prompt path still substitutes `{SOURCE_LOCALE}` / `{TARGET_LOCALE}`.
- [X] All tests pass locally (`pnpm --filter @lingo.dev/compiler test` → 229 passed).

## Visuals

N/A — no UI/UX changes (compiler and prompt only).

## Checklist

- [X] Changeset added (patch bump for `@lingo.dev/compiler`)
- [X] Tests cover business logic (selector + prompt behavior, including edge cases, not just the happy path)
- [X] No breaking changes — the default `lingo.dev` engine path is untouched, existing caches are keyed by content hash (not the prompt) so nothing is re-translated, and Spanish targets keep the exact previous example.

Closes lingodotdev/lingo.dev#2041

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed translations on cache misses potentially being primed with examples for the wrong target language.
  * Few-shot examples now match the requested locale, including Simplified and Traditional Chinese.
  * Unsupported target locales no longer receive potentially misleading examples.
* **Improvements**
  * Prompts now display readable source and target language names alongside locale codes.
  * Added coverage for Spanish, French, German, Italian, Portuguese, Dutch, Russian, Japanese, Korean, and Chinese locales.

### DIFF
--- a/.changeset/compiler-target-locale-shots.md
+++ b/.changeset/compiler-target-locale-shots.md
@@ -1,0 +1,7 @@
+---
+"@lingo.dev/compiler": patch
+---
+
+compiler: stop translations drifting to Spanish with custom OpenAI-compatible models
+
+The bundled few-shot example was always English→Spanish, so a weaker model could copy it and translate cache misses into Spanish regardless of the requested target locale (e.g. `zh-Hans`). The compiler now only sends a few-shot example whose language matches the target — curated examples ship for common languages (incl. Simplified/Traditional Chinese, Japanese, Korean) and any other target gets none rather than a wrong-language one — and it names the target language in the system prompt via `Intl.DisplayNames`. Fixes #2041.
--- a/packages/new-compiler/src/translators/lingo/prompt.test.ts
+++ b/packages/new-compiler/src/translators/lingo/prompt.test.ts
@@ -1,0 +1,29 @@
+import { describe, it, expect } from "vitest";
+import { getSystemPrompt } from "./prompt";
+
+describe("getSystemPrompt", () => {
+  it("names the source and target languages in the built-in prompt", () => {
+    const result = getSystemPrompt({ sourceLocale: "en", targetLocale: "es" });
+    expect(result).toContain("Source language (locale code): en (English)");
+    expect(result).toContain("Target language (locale code): es (Spanish)");
+  });
+
+  it("names non-Latin target languages", () => {
+    const result = getSystemPrompt({
+      sourceLocale: "en",
+      targetLocale: "zh-Hans",
+    });
+    expect(result).toContain(
+      "Target language (locale code): zh-Hans (Simplified Chinese)",
+    );
+  });
+
+  it("uses the user-defined prompt with locale substitution when provided", () => {
+    const result = getSystemPrompt({
+      sourceLocale: "en",
+      targetLocale: "es",
+      prompt: "Translate from {SOURCE_LOCALE} to {TARGET_LOCALE}.",
+    });
+    expect(result).toBe("Translate from en to es.");
+  });
+});
--- a/packages/new-compiler/src/translators/lingo/prompt.ts
+++ b/packages/new-compiler/src/translators/lingo/prompt.ts
@@ -4,6 +4,18 @@ interface PromptArguments {
   prompt?: string;
 }
 
+function getLanguageName(locale: string): string | undefined {
+  try {
+    const name = new Intl.DisplayNames(["en"], { type: "language" }).of(locale);
+    if (name && name.toLowerCase() !== locale.toLowerCase()) {
+      return name;
+    }
+  } catch {
+    // Malformed locale code — fall back to the code only.
+  }
+  return undefined;
+}
+
 export function getSystemPrompt(args: PromptArguments): string {
   // If user provided custom prompt, use it
   if (args.prompt?.trim()) {
@@ -12,6 +24,9 @@ export function getSystemPrompt(args: PromptArguments): string {
       .replace("{SOURCE_LOCALE}", args.sourceLocale)
       .replace("{TARGET_LOCALE}", args.targetLocale);
   }
+
+  const sourceName = getLanguageName(args.sourceLocale);
+  const targetName = getLanguageName(args.targetLocale);
 
   // Otherwise use built-in prompt
   return `
@@ -24,8 +39,8 @@ You replicate the meaning, intent, style, tone, and purpose of the original data
 
 ## Setup
 
-Source language (locale code): ${args.sourceLocale}
-Target language (locale code): ${args.targetLocale}
+Source language (locale code): ${args.sourceLocale}${sourceName ? ` (${sourceName})` : ""}
+Target language (locale code): ${args.targetLocale}${targetName ? ` (${targetName})` : ""}
 
 ## Guidelines
 
--- a/packages/new-compiler/src/translators/lingo/shots.test.ts
+++ b/packages/new-compiler/src/translators/lingo/shots.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect } from "vitest";
+import { getShots } from "./shots";
+
+describe("getShots", () => {
+  it("returns the Spanish example for Spanish targets", () => {
+    for (const locale of ["es", "es-ES", "es-419", "es-MX"]) {
+      const shots = getShots(locale);
+      expect(shots).toHaveLength(1);
+      expect(shots[0][1].locale).toBe("es");
+    }
+  });
+
+  it("returns the Simplified Chinese example for Simplified targets", () => {
+    for (const locale of ["zh-Hans", "zh-Hans-CN", "zh-CN", "zh-SG", "zh"]) {
+      const shots = getShots(locale);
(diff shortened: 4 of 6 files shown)


meshulga:
Hey @adityasharmawork  — thanks for this, solid work and the diagnosis is spot on. The bot closed this only because you weren't assigned to the issue, which is on our side, not yours. I've assigned you to #2041 and reopened the PR. The fix is basically there — just a couple of things before merge:

**1 (needs a change):** getShots returning [] for uncovered locales drops the only demonstration of the XML envelope — the system prompt describes the format in prose but never shows it, so weak models on nb/pl/hi etc. risk malformed XML that fails parsing. I'd teach the format in the system prompt (a small <data>/<entry> skeleton) instead of via the few-shot — then [] for uncovered locales is safe. (An en → en example would demonstrate format too, but risks nudging the model to leave text untranslated.) A test for the uncovered case would be great.

**2 (builds on 1):** Since each example now teaches the model per-language, let's ship only the ones we can confidently verify read naturally. Could you trim the table to es, fr, it, pt, de, ru and drop the rest? With change 1 in place, the dropped locales just fall back to the skeleton, so it's zero-risk — and zh-Hans is still covered by the prompt skeleton plus the language name, so the original bug stays fixed. We can add more languages back later once a native speaker has eyeballed them.

**3 (nit):** target(locale, dashboard, settings, logout, rights) — four same-typed positional strings, easy to swap silently in a non-Latin row. A named object would remove that.

Let me know if you see any of this differently.

meshulga:
Also @adityasharmawork — one process thing before this can merge: commits need to be signed (CI enforces it, see CONTRIBUTING.md). Your current one shows as Unverified. If you haven't set up commit signing yet, SSH signing is the easiest path (GitHub docs). Signing isn't retroactive, so the existing commit needs re-signing once it's set up:

git commit --amend --no-edit -S && git push --force-with-lease

(multiple commits: git rebase --exec 'git commit --amend --no-edit -S' origin/main)

meshulga (changes requested):
Just in case I will request changes with same notes I've written up there

Hey @adityasharmawork — thanks for this, solid work and the diagnosis is spot on. The bot closed this only because you weren't assigned to the issue, which is on our side, not yours. I've assigned you to https://github.com/lingodotdev/lingo.dev/issues/2041 and reopened the PR. The fix is basically there — just a couple of things before merge:

1 (needs a change): getShots returning [] for uncovered locales drops the only demonstration of the XML envelope — the system prompt describes the format in prose but never shows it, so weak models on nb/pl/hi etc. risk malformed XML that fails parsing. I'd teach the format in the system prompt (a small / skeleton) instead of via the few-shot — then [] for uncovered locales is safe. (An en → en example would demonstrate format too, but risks nudging the model to leave text untranslated.) A test for the uncovered case would be great.

2 (builds on 1): Since each example now teaches the model per-language, let's ship only the ones we can confidently verify read naturally. Could you trim the table to es, fr, it, pt, de, ru and drop the rest? With change 1 in place, the dropped locales just fall back to the skeleton, so it's zero-risk — and zh-Hans is still covered by the prompt skeleton plus the language name, so the original bug stays fixed. We can add more languages back later once a native speaker has eyeballed them.

3 (nit): target(locale, dashboard, settings, logout, rights) — four same-typed positional strings, easy to swap silently in a non-Latin row. A named object would remove that.

Let me know if you see any of this differently.

adityasharmawork:
Sure, and thanks for this advice, I will let you as soon as I am done with it!

adityasharmawork:
Sorry, I was a bit occupied with some non negotiables, I am raising a new PR fixing all issues ASAP

